PR #6, opened 2018-05-18
Update README

chrismwendt:
- I noticed that on Azure you have to specify `managed-premium` as the storage class.
- There are a few differing names for `values.yaml` that I tried to clean up


chrismwendt:
It looks like the README has been updated.

PR #8, opened 2018-05-18
add liveness probe to frontend

nicksnyder:
@beyang our old deployment had a liveness probe, did you drop this by accident?


nicksnyder:
I was mistaken, our old deployment did not have this liveness check (we just had it on sourcegraph.com, which is https, and that is where I copied this from). It does seem like we should have one though so I will change this to http.

nicksnyder:
Updated and tested on dogfood

PR #10, opened 2018-05-21
ci: Specify namespace in generate-examples

keegancsmith:
Buildkite is now running in a different namespace. It seems helm by default
uses the current namespace, which I assume it gets from the default service
account. As such this was leading to CI failing since the example templates
would have the NS set to "buildkite".


keegancsmith:
Merging in now to fix CI.

PR #22, opened 2018-05-30
update xlang-go image version

slimsag:
This updates the xlang-go image version to include a fix for the Go
language server returning poor error messages `invalid location` for
unsupported CGO files rather than a more legible `File $FILE is ignored by the build.`
error.

See https://github.com/sourcegraph/issues/issues/59


keegancsmith:
You need to run `./examples/generate-examples.sh`. I'll do it for ya and commit :)

PR #24, opened 2018-06-01
pin experimental language server image versions

ggilmore:
We now specify specific versions for each experimental language server, instead of just using the "latest" tag (which along with our default `imagePullPolicy` would prevent new versions from being pulled after the first pull).

I avoided specifying new files for each new language sever because copy-pasting the same configuration values for each new language seemed like a bad idea. Instead, I added a new `experimentalLangServers` dict to the `const` field in the values file. The key in the dict is the colloquial name for the experimental language (same as site configuration, and the image field points to the specific image on dockerhub to pull.

### DIFF
--- a/templates/xlang/experimental.yaml
+++ b/templates/xlang/experimental.yaml
@@ -1,7 +1,8 @@
 
+{{- $experimentalLangServers := .Values.const.experimentalLangServers -}}
 {{- range $index, $langserver := .Values.site.langservers -}}
 
-{{- if not (has $langserver.language (list "go" "java" "php" "python" "javascript" "typescript")) }}
+{{- if (hasKey $experimentalLangServers $langserver.language) -}}
 
 {{- if eq $index 0 | not }}
 ---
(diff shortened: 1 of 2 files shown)


beyang:
(To add a new example, add a subdirectory to `examples` that contains a `values.yaml` and `README.md`; then run `generate.examples.sh`)

ggilmore:
👍 will do

templates/xlang/experimental.yaml
```diff
@@ -45,7 +46,8 @@ spec:
         app: xlang-{{ $langserver.language }}
     spec:
       containers:
-      - image: sourcegraph/codeintel-{{ default $langserver.language (index (dict "cs" "csharp" "dockerfile" "docker") $langserver.language) }}
+      {{ $imageName := (pluck "image" (pluck $langserver.language $experimentalLangServers | first) | first) -}}
+      - image: {{ $imageName }}
```
chrismwendt: I'm not 100% sure about the syntax, but it looks like this is just grabbing a nested field. Could you use [`index`](https://godoc.org/text/template)? Something like this:

```
- image: {{ index $experimentalLangServers $langserver.language image }}
```

ggilmore:
@beyang PTAL

PR #29, opened 2018-06-04
remove `test` script and add test-cases/custom-ls-address

beyang:
@chrismwendt This is what I had in mind for the test case from the discussion this morning. This PR removes the `test` script and replaces it with a directory containing a `values.yaml` file that describes the config we want to test (in this case, specifying a custom LS address). This has the following benefits over the `test` script:
* We get full kubeval validation over the generated config.
* The generated config is easier vendored for easier debugability.
* It doesn't really affect searchability at all, since we already vendor multiple generated configs in the `examples` directory.
* You still notice when a change to the chart would cause a regression to the generated config. 
* You'll notice an undesired change even if it's not explicitly checked for by a test case.
* It's easier to update the test case if need be (`git add` vs modifying the `test` script)

The downside is that you have to eyeball the change to the test case to determine if it's legit. I think this is pretty low cost and worth it to catch potentially unchecked for changes to the config.

The `test-cases` directory essentially functions the same as the `examples` directory, but is separate because not all test cases are ones we want to present to the end user as examples to emulate.


chrismwendt:
Seeing a diff is way nicer, sounds like an improvement on many fronts :+1: :sparkles:

I ran it just now and I see some docker images changed:

```diff
diff --git a/test-cases/custom-ls-address/generated/sourcegraph/templates/frontend/sourcegraph-frontend.Deployment.yaml b/test-cases/custom-ls-address/generated/sourcegraph/templates/frontend/sourcegraph-frontend.Deployment.yaml
index d990f9e..8c9667e 100644
--- a/test-cases/custom-ls-address/generated/sourcegraph/templates/frontend/sourcegraph-frontend.Deployment.yaml
+++ b/test-cases/custom-ls-address/generated/sourcegraph/templates/frontend/sourcegraph-frontend.Deployment.yaml
@@ -72,7 +72,7 @@ spec:
           value: k8s+http://symbols:3184
         - name: ZOEKT_HOST
           value: indexed-search:80
-        image: docker.sourcegraph.com/frontend:16305_2018-06-01_d377b05
+        image: docker.sourcegraph.com/frontend:16491_2018-06-05_b9f9e8a
         livenessProbe:
           httpGet:
             path: /healthz
diff --git a/test-cases/custom-ls-address/generated/sourcegraph/templates/gitserver/gitserver.Deployment.yaml b/test-cases/custom-ls-address/generated/sourcegraph/templates/gitserver/gitserver.Deployment.yaml
index c252732..884ba4f 100644
--- a/test-cases/custom-ls-address/generated/sourcegraph/templates/gitserver/gitserver.Deployment.yaml
+++ b/test-cases/custom-ls-address/generated/sourcegraph/templates/gitserver/gitserver.Deployment.yaml
@@ -34,7 +34,7 @@ spec:
           value: :6060
         - name: SRC_REPOS_DIR
           value: /data/repos
-        image: docker.sourcegraph.com/gitserver:16098_2018-05-29_e34461d
+        image: docker.sourcegraph.com/gitserver:16476_2018-06-05_e786c66
         livenessProbe:
           initialDelaySeconds: 5
           tcpSocket:
diff --git a/test-cases/custom-ls-address/generated/sourcegraph/templates/repo-updater/repo-updater.Deployment.yaml b/test-cases/custom-ls-address/generated/sourcegraph/templates/repo-updater/repo-updater.Deployment.yaml
index 58772f5..7a9447e 100644
--- a/test-cases/custom-ls-address/generated/sourcegraph/templates/repo-updater/repo-updater.Deployment.yaml
+++ b/test-cases/custom-ls-address/generated/sourcegraph/templates/repo-updater/repo-updater.Deployment.yaml
@@ -33,7 +33,7 @@ spec:
           value: gitserver-1:3178
         - name: SRC_PROF_HTTP
           value: :6060
-        image: docker.sourcegraph.com/repo-updater:14230_2018-04-13_770f492
+        image: docker.sourcegraph.com/repo-updater:16492_2018-06-05_b9f9e8a
         name: repo-updater
         ports:
         - containerPort: 3182
```

beyang:
Thanks for noting, fixed in def66d870b1c13dd35a76aaddd9a714a7c88ebd7

PR #30, opened 2018-06-05
bump repo-updater

seebs:
a recent change to the gitserver client affected frontend and repo-updater.

note: it also affected frontend, i think, but i don't know the magic for updating deploy-sourcegraph, or whether magic is needed? sending this to get the buildkite test and because readme.dev.md says so.


KattMingMing:
Released in https://github.com/sourcegraph/deploy-sourcegraph/pull/31

PR #39, opened 2018-07-05
documentation improvements

seebs:
Hypothesis: if I get stalled or confused by things, other people might too, so add clarifications to the documentation.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -63,7 +63,6 @@ sure you have [configured `kubectl` to access your cluster](https://kubernetes.i
      underlying disk type. For more info, see the section below on "creating a storage class
      manually".
 
-
 1. Install the Helm chart to your cluster:
 
    ```bash
--- a/docs/install.md
+++ b/docs/install.md
@@ -43,7 +43,9 @@ sure you have [configured `kubectl` to access your cluster](https://kubernetes.i
        create: {none,aws,gcp}
        name: $NAME
        zone: $ZONE
-   site: {}
+   site: {
+     "auth.providers": [{"type": "builtin", "allowSignup": false}],
+   }
    ```
 
    - If using Google Cloud, set `cluster.storageClass.create` to `gcp` and
@@ -60,6 +62,14 @@ sure you have [configured `kubectl` to access your cluster](https://kubernetes.i
      Kubernetes cluster with name "default". We recommend that the storage class use SSDs as the
      underlying disk type. For more info, see the section below on "creating a storage class
      manually".
+
+1. Examine the sample `values.yaml` in the `deploy-sourcegraph` repo for additional starting
+   configuration values. Do not copy fields in if you don't need to change them; the values
+   in your user-provided file are merged in on top of this file. The existing resource
+   allocations are probably a reasonable starting point.
+
+   If you find you've made errors, you can update `values.yaml` using `helm upgrade`; for
+   more information, read about [upgrading Sourcegraph Data Center](upgrade.md).
 
 1. Install the Helm chart to your cluster:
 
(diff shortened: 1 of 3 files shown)


docs/install.md
```diff
@@ -61,6 +63,14 @@ sure you have [configured `kubectl` to access your cluster](https://kubernetes.i
      underlying disk type. For more info, see the section below on "creating a storage class
      manually".
 
+1. Examine the sample `values.yaml` in the `deploy-sourcegraph` repo for additional starting
+   configuration values. Do not copy fields in if you don't need to change them; the values
+   in your user-provided file are merged in on top of this file. The existing resource
+   allocations are probably a reasonable starting point.
+
+   If you find you've made errors, you can update `values.yaml` using `helm upgrade`; for
```
beyang: This is inaccurate. `helm upgrade` upgrades the running k8s cluster; it does not change `values.yaml`.

docs/install.md
```diff
@@ -61,6 +63,14 @@ sure you have [configured `kubectl` to access your cluster](https://kubernetes.i
      underlying disk type. For more info, see the section below on "creating a storage class
      manually".
 
+1. Examine the sample `values.yaml` in the `deploy-sourcegraph` repo for additional starting
```
beyang: How about moving this section to the end. I'd prefer that folks get something up and running first before customizing config. It also segues nicer into update.md that way.

docs/install.md
```diff
@@ -72,6 +82,15 @@ sure you have [configured `kubectl` to access your cluster](https://kubernetes.i
 1. Confirm that your deployment is launching by running `kubectl get pods`. If pods get stuck in `Pending` status, run
    `kubectl get pv` to check if the necessary volumes have been provisioned (you should see at least 4). Google Cloud
    Platform users may need to [request an increase in storage quota](https://cloud.google.com/compute/quotas).
+   If you have enough volumes, but pods are still stuck, `kubectl cluster-info dump` (very verbose, dump it to a file)
+   may have more information; for instance:
+   ```
+    "Reason": "FailedScheduling",
```
beyang: This seems like it's addressing a very specific issue that may not come up for many installers. Could you move it to a separate "Troubleshooting" section at the end of the document?

Also, as this documentation is public-facing, could we be stricter on grammar, punctuation, and writing style. For instance,


You can run `kubectl cluster-info dump > dump.txt` to obtain a verbose dump of the cluster state, which can be helpful for debugging issues. Look for warning messages like the following:

```
"Reason": "FailedScheduling",
"Message": "0/3 nodes are available: 1 Insufficient memory, 3 Insufficient cpu.",
```

docs/install.md
```diff
@@ -80,7 +99,8 @@ sure you have [configured `kubectl` to access your cluster](https://kubernetes.i
    (see
    [AWS Security Group rules](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_SecurityGroups.html),
    [Google Cloud Platform Firewall rules](https://cloud.google.com/compute/docs/vpc/using-firewalls)). Sourcegraph
-   should then be accessible at `$EXTERNAL_ADDR_OF_YOUR_NODE:30080`. For production environments, we recommend using
+   should then be accessible at `$EXTERNAL_ADDR_OF_YOUR_NODE:30080`. If you have multiple nodes, all the addresses
```
beyang: Change to "should then be accessible at `$EXTERNAL_NODE_ADDR:30080`, using the external address of *any* of the nodes in the cluster."

No need to mention the implementation detail of `kube-proxy` here.

docs/install.md
seebs: Er. yes. "Deploy updated values.yaml to the cluster" or something.

docs/install.md
seebs: I think when I first wrote it, I wasn't aware of how to update the config, but also my initial pass didn't have an auth thing, so looking through values.yaml to find out I needed one was important to do early on. But yeah, it could move.

docs/install.md
seebs: good catch, and yes, this belongs in the troubleshooting document, which already exists as a separate document; the troubleshooting "section" in this one is just a link there. Which is reasonable.

seebs:
good catches. revised with new commit.

seebs:
@beyang, ptal; I think I've addressed the comments, but I should probably verify with someone who isn't me. :)

docs/install.md
```diff
@@ -69,24 +72,33 @@ sure you have [configured `kubectl` to access your cluster](https://kubernetes.i
...
+   [Google Cloud Platform Firewall rules](https://cloud.google.com/compute/docs/vpc/using-firewalls)).
+   Sourcegraph should then be accessible at `$EXTERNAL_ADDR:30080`, where `$EXTERNAL_ADDR` is the
+   address of *any* node in the cluster. For production environments, we recommend using
    an [Internet Gateway](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Internet_Gateway.html) (or
    equivalent) and configuring a load balancer in Kubernetes.
 
 You will now see the Sourcegraph setup page when you visit the address of your instance. If you made your instance
 accessible on the public Internet, make sure you secure it before adding your private repositories.
 
+## Additional configuration
```
beyang: This seems redundant to the "Additional configuration" section below?

PR #41, opened 2018-07-10
Bump versions to use new redis envvar behaviour

keegancsmith:
We have switched to use the more consistent `REDIS_CACHE_ENDPOINT` and `REDIS_STORE_ENDPOINT` to specify the addresses of our two redis instances. This PR bumps all the image versions to read the new environment variable. Additionally we remove specifying the environment variables since the defaults work in kubernetes.

Part of https://github.com/sourcegraph/sourcegraph/issues/12221

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -114,13 +114,11 @@
 {{- $_ := set .envVars "LSP_PROXY" "lsp-proxy:4388" -}}
 
 {{- $_ := set .envVars "PUBLIC_REPO_REDIRECTS" "\"true\"" -}}
-{{- $_ := set .envVars "REDIS_MASTER_ENDPOINT" "redis-cache:6379" -}}
 {{- $_ := set .envVars "SEARCHER_URL" "k8s+http://searcher:3181" -}}
 {{- $_ := set .envVars "SRC_GIT_SERVERS" (include "gitservers" .) -}}
 {{- $_ := set .envVars "SRC_INDEXER" "indexer:3179" -}}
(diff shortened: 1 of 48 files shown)


nicksnyder:
I think this is fine to merge and then cut a prerelease version so we can deploy it to dogfood for testing.

templates/_helpers.tpl
```diff
@@ -116,14 +116,14 @@
 
 {{- $_ := set .envVars "PUBLIC_REPO_REDIRECTS" "\"true\"" -}}
 {{- $_ := set .envVars "QUERY_RUNNER_URL" "http://query-runner" -}}
-{{- $_ := set .envVars "REDIS_MASTER_ENDPOINT" "redis-cache:6379" -}}
+{{- $_ := set .envVars "REDIS_CACHE_ENDPOINT" "redis-cache:6379" -}}
+{{- $_ := set .envVars "REDIS_STORE_ENDPOINT" "redis-store:6379" -}}
```
nicksnyder: Can we just not set these and rely on the default behavior that now exists in the code?

slimsag:
I agree with Nick, I think this is fine.

I will also be working on sourcegraph/sourcegraph#10767 this week, which will conflict with this & ultimately I'd want this in before closing that issue as well I think, so it'd be nice to do this sooner rather than later.

keegancsmith:
We can actually just not specify the environment variables. In the new and old code we use `redis-cache:6379` and `redis-store:6379`. I'll update this PR to just bump the versions and remove the env vars. I assume we still want to bump the versions to have the nicer envvars for the issue @slimsag is working on.

PR #51, opened 2018-07-11
remove ZOEKT_HOST and rely on default behavior

nicksnyder:
- [x] update image after https://github.com/sourcegraph/sourcegraph/pull/12278 is merged


nicksnyder:
Yes, that is the TODO in the description

PR #53, opened 2018-07-12
remove SRC_PROF_HTTP and rely on default behavior

nicksnyder:
Need to bump images:
- [x] sourcegraph-frontend
- [x] github-proxy
- [x] gitserver
- [x] indexer
- [x] lsp-proxy
- [x] query-runner
- [x] repo-updater
- [x] searcher
- [x] symbols
- [x] xlang-go

Other images that got updated
- xlang-typescript
- xlang-php


keegancsmith:
You can use `update_docker_image_versions.py` to sync the image versions to what we are running on dogfood.

PR #56, opened 2018-07-13
Make specifying custom PostgreSQL / Redis servers a first class feature

slimsag:
This PR makes specifying a custom/external PostgreSQL and Redis server a first-class feature.

In short, you can now specify the `customPostgres` or `customRedis` section of the `cluster` in order to make all services use the specified DB. See `values.yaml` for examples.

Appropriate tests are also added, but please review carefully as I want to make certain this change is correct :)

Adding proper docs for this on the website (for both Server and Data Center) will be done in a follow-up change once this is merged.


slimsag:
Good point, thank you for mentioning that! Not nit-picky at all.

I will update in the morning :)

nicksnyder:
I agree with removing `custom` prefix.

If this is specified, wouldn't we also want to exclude the existing pgsql and redis templates from being generated?

@slimsag are you planning to ship this to Netflix before the end of the iteration?

slimsag:
> If this is specified, wouldn't we also want to exclude the existing pgsql and redis templates from being generated?

Good point, I'll look into it.

> @slimsag are you planning to ship this to $CUSTOMER before the end of the iteration?

I think we can wait to ship the Data Center change here until the end of the iteration, but some other parts (Server docs) will need to go out sooner.

slimsag:
@nicksnyder given our convo just now, and the fact that this is more of a backup, I'm gonna punt on excluding the existing pgsql and redis templates from being generated since I didn't see an easy way to tackle this and they are mostly harmless anyway. We can just tackle this in the kustomize implementation instead.

PTAL

slimsag:
Thanks for the pointer! Did that & will merge once CI passes.

PR #60, opened 2018-08-01
correct the outdated syntect_server version

slimsag:
The commit https://github.com/sourcegraph/deploy-sourcegraph/commit/b74e58681aeb4188df62ae6ce4932b50567664ce updated the syntect_server version using the new `update_docker_image_versions.py` script. However, our dogfood instance does not have automatic deployments of syntect_server (it is special in this regard), and hence using the version from there is incorrect as it can be outdated.

The syntect_server version is now hard-coded in `update_docker_image_versions.py`, and the `values.yaml` file has been manually adjusted to use the latest `syntect_server` version once again (done manually to avoid updating other image versions for a patch release).

This fixes a regression where e.g. Java files would be improperly highlighted. See internal issue: sourcegraph/sourcegraph#12537


slimsag:
Merging now so that this is in the next release ASAP.

PR #61, opened 2018-08-01
2.10.0 release

ggilmore:
I ran:

```shell
./update-docker-image-versions.py
./generate.sh
``` 

and this was the result. Once this is approved, it will still need to be tested on dogfood.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -6,36 +6,36 @@
 # const defines configuration constants that generally should not be changed unless testing out experimental or pre-release features.
 const:
   frontend:
-    image: docker.sourcegraph.com/frontend:18282_2018-07-13_d76e9c6
+    image: docker.sourcegraph.com/frontend:18969_2018-07-31_6ae5417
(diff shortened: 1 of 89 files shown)


values.yaml
```diff
@@ -6,36 +6,36 @@
 # const defines configuration constants that generally should not be changed unless testing out experimental or pre-release features.
 const:
   frontend:
-    image: docker.sourcegraph.com/frontend:18282_2018-07-13_d76e9c6
+    image: docker.sourcegraph.com/frontend:18969_2018-07-31_6ae5417
   searcher:
-    image: docker.sourcegraph.com/searcher:18235_2018-07-12_21a0f96
+    image: docker.sourcegraph.com/searcher:18570_2018-07-19_56d5400_patch
```
nicksnyder: Using a patch here seems wrong.

For a minor release, we should be updating the tags to point to the tagged version.

PR #63, opened 2018-08-01
Update dev readme with clarified instructons for tagging and testing on new clusters

ggilmore:


README.dev.md
```diff
@@ -2,24 +2,19 @@
...
-* Test this release on a fresh cluster cutting a real release. If this release includes changes not
-  yet deployed to dogfood (generally if it includes k8s config changes, rather than merely Docker
-  image changes), cut a test release (`LATEST=false ./release $TEST_VERSION`). Then go to the
-  `infrastructure` repository and follow the instructions in `datacenter/README.md` to update
-  dogfood to the test release.
-* Run `./release.sh $VERSION`.
-  * If this is a patch version to a previous major/minor version, use `LATEST=false ./release.sh
-    $VERSION`.
-  * If this is a release candidate, run `LATEST=false ./release {$VERSION}-pre${N}` where `N` starts
+- Make the desired changes to this repository. Most commonly, this involves updating the Docker image versions in `values.yaml` to match the tagged version you are releasing.
```
slimsag: Given we're updating these manually @ggilmore @nicksnyder -- when do our language server images get updated? I see that xlanGo, xlangJava, xlangTypescriptJavascript are not tagged with 2.10 and I worry we will end up forgetting to update these because of it.

values.yaml
```diff
@@ -25,6 +23,8 @@ const:
     image: docker.sourcegraph.com/query-runner:2.10.0
   repoUpdater:
     image: docker.sourcegraph.com/repo-updater:2.10.0
+  indexedSearch:
+    image: docker.sourcegraph.com/zoekt:18-05-30_3d2275e
```
slimsag: Why isn't this tagged, as well?

slimsag:
I'm worried that you deleted our `update_docker_image_versions.py` script @nicksnyder while a large portion of our images remain untagged with a release version number. I am afraid that we will regress as we've done in the past by not properly updating these image versions.

values.yaml
ggilmore: Our zoekt fork is not update on the same release-schedule as all of the other images above it in this list.

slimsag: Oops, I missed that this was zoekt and just read `indexedSearch`. That makes sense for this one. Thanks!

README.dev.md
ggilmore: Right now all of our language servers are updated on a different release cycle than all of the other data center components. I can't think of a time where we made a breaking change in the language server (i.e. you can mix and match versions). I'm not opposed to the idea of versioning it along side the other components though.

nicksnyder:
I deleted the script because its existence was actively harmful since it encouraged doing the wrong thing. I agree that ideally we can/should script the other image tags, but that requires a little bit more effort that has not been done yet.

README.dev.md
slimsag: I am not worried about breaking changes. I am worried that data center users will not receive the latest updates to our language servers simply because we will forget to update the images here. This happened very often before, which was why @beyang had added the `update_docker_image_versions.py` script. But now that script is gone / not a part of our release process here, meaning it is trivial once again to forget to update these as part of cutting a release.

slimsag: (To be clear, I'm not saying we need to version them, I'm just saying we need this release process to *always* update them. Maybe that's as simple as adding instructions here to go to our CI and update each langserver version with the latest found there.)

slimsag:
Please see my comment here: https://github.com/sourcegraph/deploy-sourcegraph/pull/63#discussion_r207067174

I'm okay with the script being deleted, and with the process not being automated. But we should make sure the langserver images get updated regardless.

ggilmore:
@slimsag https://github.com/sourcegraph/deploy-sourcegraph/pull/63/commits/7fd60bcc01f202ab998c23b745769c1798d0a764

README.dev.md
```diff
@@ -2,7 +2,9 @@
 
 ## Cutting a release
 
-- Make the desired changes to this repository. Most commonly, this involves updating the Docker image versions in `values.yaml` to match the tagged version you are releasing.
+- Make the desired changes to this repository. Most commonly, this involves updating the Docker image versions in `values.yaml` to match the tagged version you are releasing. For images like language servers, you can look at our
```
slimsag: s/you can/you should/g

PR #65, opened 2018-08-08
(DON'T MERGE YET) Gitserver Stateful Sets

ggilmore:
This PR changes `gitserver`'s deployment to a StatefulSet instead. 

https://github.com/sourcegraph/deploy-sourcegraph/pull/65/files#diff-18a3c769d8a8bfa4a665c9ec3cffa8a6 is probably the most relevant diff to pay attention to. 

--- 

🚨 This change requires some manual steps before upgrading: see https://github.com/sourcegraph/deploy-sourcegraph/blob/7f4f9eb78eb797e3e040b6afe69f77d9cba53f70/docs/stateful-set-migrate.md. 🚨

How I tested this: 

1. Setup a `2.10` GCP cluster with the `values.yaml` below by running:

```shell
kubectl create serviceaccount --namespace kube-system tiller

kubectl create clusterrolebinding tiller --clusterrole=cluster-admin --serviceaccount=kube-system:tiller

helm init --service-account tiller

helm install --name sourcegraph -f values.yaml https://github.com/sourcegraph/datacenter/archive/latest.tar.gz
```

<details>

`values.yaml`:

```yaml
cluster:
  gitserver:
    # Number of shards for caching git data. If you change this, you should also change indexedSearch.diskSize.
    shards: 2
    # Size of disk of each gitserver shard. If you change this, you should also change indexedSearch.diskSize.
    diskSize: 400Gi
  indexedSearch:
    # The size of disk to used for search indexes (this should typically be gitserver.diskSize multipled by gitserver.shards)
    diskSize: 400Gi
  storageClass:
    create: gcp
    zone: us-central1-a


site: {
  # Publicly accessible URL to web app (e.g., what you type into your browser).
  "appURL": "http://localhost:3080",

  # Port exposed on all nodes directing HTTP traffic to sourcegraph-frontend pod(s).
  "httpNodePort": 30080,

  # # Port exposed on all nodes directing HTTPS traffic to sourcegraph-frontend pod(s). Uncomment this
  # # to allow incoming HTTPS traffic.
  # httpsNodePort: 30081

  # # The authentication provider to use for identifying and signing in users. Only one entry is supported.
  # #
  # # The builtin auth provider with signup disallowed (shown below) means that after the initial site admin signs in, all other users must be invited.
  # #
  "auth.providers": [{"type": "builtin", "allowSignup": false}],

  "github": [
    {
      "initialRepositoryEnablement": true,
      "repos": [
        facebook/react,
        golang/go,
        sqs/spans,
        tortis/sleepy,
        slimsag/gfx,
        golangbox/gobox
      ],
      "token": "<REDACTED>",
      "url": "https://github.com"
    },
  ], 

  "maxReposToSearch": 999999
}
```

</details>

2. Sign up, make a few searches, wait until all repositories are cloned

3. Run through the steps described in https://github.com/sourcegraph/deploy-sourcegraph/blob/7f4f9eb78eb797e3e040b6afe69f77d9cba53f70/docs/stateful-set-migrate.md

- the volumes associated with `gitserver-1,2` had their status go from `bound -> released -> available`

- `gitserver`'s claims were deleted 

4. With the `stateful` branch in `deploy-sourcegraph` checked out, run `helm upgrade sourcegraph -f values.yaml $DEPLOY_SOURCEGRAPH_ROOT`

- The same volumes above seem to be bound again (and no new ones were provisioned) 
- No additional clones seem to have been done


templates/gitserver/gitserver.Service.yaml
```diff
@@ -7,31 +7,9 @@ metadata:
 spec:
   clusterIP: None
   ports:
-  - name: unused
-    port: 10811
-    targetPort: 10811
+    - name: unused
```
nicksnyder: was this broken before?

ggilmore: I'm not sure what you mean. If it's just the indentation, then VSCode/prettier formats it like this.

nicksnyder: Yes, I mean the formatting. Looks like it is equivalent in yaml, I just wanted to double check that the whitespace wasn't significant here

templates/_helpers.tpl
```diff
@@ -145,16 +145,16 @@
 {{/* --------------- START OF TEMPLATE ------------- */}}
 
 {{- define "gitservers" -}}
-{{- include "joinFmt" (dict "count" (int (default 1 .Values.cluster.gitserver.shards)) "fmt" "gitserver-%d:3178" "sep" " ") -}}
+{{- include "joinFmt" (dict "count" (int (default 1 .Values.cluster.gitserver.shards)) "fmt" "gitserver-%d.gitserver.default.svc.cluster.local:3178" "sep" " ") -}}
```
ggilmore: Before, this function returned 

```
"gitserver-1:3178 gitserver-2:3178 gitserver-3:3178 ..."
```

it now returns: 
```
"gitserver-0.gitserver.default.svc.cluster.local:3178 gitserver-1.gitserver.default.svc.cluster.local:3178 gitserver-2.gitserver.default.svc.cluster.local:3178 ..."
```


so that pods can connect to it as described in https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-identity

It does seem strange/brittle that I need to hardcode the namespace (`default`) here though.

templates/gitserver/gitserver.StatefulSet.yaml
```diff
@@ -6,29 +6,25 @@
...
 metadata:
   annotations:
     description: Stores clones of repositories to perform Git operations.
-  name: gitserver-{{ add1 $gitserverIndex }}
+  name: gitserver
 spec:
-  minReadySeconds: 10
-  replicas: 1
+  serviceName: gitserver
+  selector:
```
ggilmore: I needed to add this selector as described by https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-selector

templates/gitserver/gitserver.StatefulSet.yaml
```diff
@@ -6,29 +6,25 @@
...
-kind: Deployment
+apiVersion: apps/v1
+kind: StatefulSet
 metadata:
   annotations:
     description: Stores clones of repositories to perform Git operations.
-  name: gitserver-{{ add1 $gitserverIndex }}
+  name: gitserver
 spec:
-  minReadySeconds: 10
```
ggilmore: `minReadySeconds` is not supported in stateful sets https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-selector

templates/gitserver/gitserver.StatefulSet.yaml
```diff
@@ -6,29 +6,25 @@
...
-  replicas: 1
+  serviceName: gitserver
+  selector:
+    matchLabels:
+      app: gitserver
+  replicas: {{ required ".Values.cluster.gitserver.shards required" .Values.cluster.gitserver.shards | int }}
   revisionHistoryLimit: 10
-  strategy:
-    rollingUpdate:
```
ggilmore: `maxSurge` and `maxUnavailable` aren't supported for stateful sets https://godoc.org/k8s.io/api/apps/v1#RollingUpdateStatefulSetStrategy

templates/gitserver/gitserver.StatefulSet.yaml
```diff
@@ -68,12 +64,20 @@ spec:
           name: config-file
         name: sg-config
       - name: repos
-        persistentVolumeClaim:
```
ggilmore: Specifying `persistentVolumeClaim` here is now necessary, StatefulSet will refer to the claim below that has the same name as the volume (`repos`)

docs/stateful-set-migrate.md
```diff
@@ -0,0 +1,37 @@
...
+3. Delete the old `gitserver`'s persistent volume claims
+
+```bash
+kubectl get pvc -o json | jq --raw-output ".items | map(select(.metadata.name | contains(\"gitserver-\"))) | .[] | \"kubectl delete pvc \\(.metadata.name)\"" | bash
+```
+
+4. Update `gitserver`'s persistent volumes so they can be reused by the new StatefulSet
+
+```bash
+kubectl get pv -o json | jq --raw-output ".items | map(select(.spec.claimRef.name | contains(\"gitserver-\"))) | .[] | \"kubectl patch pv -p '{\\\"spec\\\":{\\\"claimRef\\\":{\\\"uid\\\":null,\\\"name\\\":\\\"repos-gitserver-\\(.spec.claimRef.name | ltrimstr(\"gitserver-\") | tonumber - 1)\\\"}}}' \\(.metadata.name)\"" | bash
```
ggilmore: This line transforms `pvc` names like:

```
gitserver-1, gitserver-2, … -> repos-gitserver-0, repos-gitserver-1, …
``

(`repos` is the name of the claim in the statefulset volumeclaimtemplate)

File: templates/gitserver/gitserver.Service.yaml (hunk 1 above)
Comment (ggilmore) [reply templates/gitserver/gitserver.Service.yaml]: I assume that it wasn't auto-formatted previously because of the syntax-errors introduced by the templating logic.

**Comment by ggilmore** [comment]
@keegancsmith @nicksnyder PTAL, I would appreciate someone else testing this / giving it another pass

File: docs/stateful-set-migrate.md (hunk 7 above)
Comment (nicksnyder) [reply docs/stateful-set-migrate.md]: I would note this in the documentation itself. Something like "the old PVC names look like `gitserver-1, gitserver-2` and the new ones look like `repos-gitserver-0, repos-gitserver-1`"

File: docs/stateful-set-migrate.md [hunk 8]
```diff
@@ -0,0 +1,37 @@
...
+```
+
+5. (**downtime ends here**) [Proceed with the normal upgrade steps](./update.md)
```
nicksnyder: Little bit unclear that downtime ends after the upgrade. Perhaps move the downtime note to the end of the line or put it in step 6

templates/_helpers.tpl
nicksnyder: Do you really need the fully qualified domain here? In https://github.com/sourcegraph/sourcegraph/pull/12418/files seems to imply that the `gitserver-0.gitserver` part should be sufficient. It is also worth trying to see if `gitserver-0` is sufficient.

ggilmore: I think you're right in that I don't need the entire domain name.

screen shot 2018-08-09 at 1 53 18 pm

Just `gitserver-0` didn't work when I tried it earlier. I don't think there are individual services for each replica - just the governing one.

templates/gitserver/gitserver.StatefulSet.yaml
```diff
@@ -6,29 +6,25 @@
 {{- $_ := set $envVars "ORIGIN_MAP" $global.Values.site.gitOriginMap -}}
 {{- include "collectEnv" (list $envVars .Values.cluster.gitserver.containers.gitserver.env) -}}
 
-{{- range $index, $gitserverIndex := until (required ".Values.cluster.gitserver.shards required" .Values.cluster.gitserver.shards | int) }}
-{{- if eq $index 0 | not }}
----
-{{ end -}}
-apiVersion: extensions/v1beta1
-kind: Deployment
+apiVersion: apps/v1
```
nicksnyder: This requires Kubernetes 1.9+
If we used `apps/v1beta2` the minimum requirement would be 1.8+
If we used `apps/v1beta1` the minimum requirement would be 1.5+

I tested with `apps/v1beta1` on Kubernetes 1.8.10-gke.0 (which is the lowest GCP allows) and it seemed to work.

We should see what version of Kubernetes our customers are already using, choose an appropriate value here, and document the minimum version of Kubernetes that we require.

docs/stateful-set-migrate.md
```diff
@@ -0,0 +1,39 @@
...
+    ```bash
+    kubectl get pvc -o json | jq --raw-output ".items | map(select(.metadata.name | contains(\"gitserver-\"))) | .[] | \"kubectl delete pvc \\(.metadata.name)\"" | bash
+    ```
+
+4. Update `gitserver`'s persistent volumes so they can be reused by the new StatefulSet
+
+    This step transforms the the old `claimRef.name`s that looked like `gitserver-1, gitserver-2, ...` into `repos-gitserver-0, repos-gitserver-1, ...`.
+
+    ```bash
+    kubectl get pv -o json | jq --raw-output ".items | map(select(.spec.claimRef.name | contains(\"gitserver-\"))) | .[] | \"kubectl patch pv -p '{\\\"spec\\\":{\\\"claimRef\\\":{\\\"uid\\\":null,\\\"name\\\":\\\"repos-gitserver-\\(.spec.claimRef.name | ltrimstr(\"gitserver-\") | tonumber - 1)\\\"}}}' \\(.metadata.name)\"" | bash
```
keegancsmith: Avoiding JQ (if you want to)
```bash
kubectl get pv --no-headers -o=custom-columns=CLAIM:.spec.claimRef.name,NAME:.metadata.name \
 | awk '/^gitserver/ { $1="repos-gitserver-" substr($1,length($1),1)-1; print }' \
 | awk '{ $1="{\"spec\":{\"claimRef\":{\"uid\":null,\"name\":\"" $1 "\"}}}"; print }' \
 | xargs -n 2 kubectl patch pv -p
```

templates/_helpers.tpl
```diff
@@ -145,16 +145,16 @@
 {{/* --------------- START OF TEMPLATE ------------- */}}
 
 {{- define "gitservers" -}}
-{{- include "joinFmt" (dict "count" (int (default 1 .Values.cluster.gitserver.shards)) "fmt" "gitserver-%d:3178" "sep" " ") -}}
+{{- include "joinFmt" (dict "count" (int (default 1 .Values.cluster.gitserver.shards)) "fmt" "gitserver-%d.gitserver:3178" "sep" " ") -}}
```
keegancsmith: default should be 0 not 1

templates/gitserver/gitserver.Service.yaml
keegancsmith: this is different indentation to what kubernetes pretty prints out. but if CI is happy, this should be good (since we validate with kubeval)

templates/gitserver/gitserver.StatefulSet.yaml
```diff
@@ -68,12 +64,20 @@ spec:
...
+  - metadata:
+      annotations:
+        volume.beta.kubernetes.io/storage-class: {{ $global.Values.cluster.storageClass.name }}
+      name: repos
+    spec:
+      accessModes:
+      - ReadWriteOnce
+      resources:
+        requests:
+          storage: {{ required ".Values.cluster.gitserver.diskSize is required" $global.Values.cluster.gitserver.diskSize }}
```
keegancsmith: I know we have some custom volume stuff. Just wondering if this looks good to you @beyang

templates/_helpers.tpl
ggilmore: You should at least have 1 shard though

keegancsmith: if not specified, don't you want it to print `gitserver-0.gitserver:3178`?

templates/gitserver/gitserver.StatefulSet.yaml
ggilmore: Not sure how relevant all the above is, but those seem to be the changes that affect stateful sets

templates/_helpers.tpl
ggilmore: It does? See one of the examples: https://github.com/sourcegraph/deploy-sourcegraph/blob/stateful/examples/gcp/generated/sourcegraph/templates/frontend/sourcegraph-frontend.Deployment.yaml#L49-L50

ggilmore: I changed `joinFmt` below to count from `[0, n)` instead of `[1, n]`. No other function uses it, so this seemed safe to do. Having the `count` represent a default of a single shard versus the `0`th shard made sense to me

templates/gitserver/gitserver.StatefulSet.yaml
nicksnyder: We should prefer `apps/v1` unless we find out that one of our existing DC customers is using less that 1.9

templates/gitserver/gitserver.StatefulSet.yaml
nicksnyder: Looking at the diff, it doesn't look like we regressed on any volume stuff.

docs/stateful-set-migrate.md
```diff
@@ -0,0 +1,39 @@
...
+
+In order be able to re-use your existing `gitserver`'s persistent volumes with [StatefulSets](https://kubernetes.io/docs/tutorials/stateful-application/basic-stateful-set/) (so that you can avoid re-cloning existing repositories), your cluster's administrator will need to run the following manual steps before upgrading to `2.11` ([click here to read more about why this is necessary](https://github.com/kubernetes/kubernetes/issues/48609#issuecomment-314066616)):
+
+## Steps
+
+_Note that these steps will cause a small amount of unavoidable downtime._
+
+_The following steps assume that you have [jq](https://stedolan.github.io/jq/) installed._
+
+1. Make sure that the persistent volumes for your existing `gitserver` deployment are marked as `retained`
```
beyang: The command looks like it updates the PVs rather than just displaying their status, so a better sentence would be "Set the Reclaim Policy for ... to Retain."

templates/gitserver/gitserver.StatefulSet.yaml
beyang: We use manually provisioned volumes in prod, but that doesn't use this chart. AFAIK no customers use manual volumes, as the docs tells you to create a storage class.

Any volume customization would therefore be through the storage class. As long as this matches the old PVC (which this does), we're okay.

PR #74, opened 2018-08-31
version 2.11.0-helm

ggilmore:
I have not tested this yet


nicksnyder:
I was thinking we should do this after we merge the yaml branch, not before. That way we dogfood the new instructions.

ggilmore:
@nicksnyder done

PR #79, opened 2018-09-05
add namespace:default to cluster role binding subjects

ggilmore:


configure/prometheus/prometheus.ClusterRoleBinding.yaml
```diff
@@ -11,3 +11,4 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: prometheus
+  namespace: default
```
nicksnyder: keep newline at end of file

configure/ssd/pod-tmp-gc.ClusterRoleBinding.yaml
```diff
@@ -11,3 +11,4 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: pod-tmp-gc
+  namespace: default
```
nicksnyder: keep newline at end of file

ggilmore:
@nicksnyder PTAL

PR #81, opened 2018-09-06
add docs about configuring gitserver's replica count before initally deploying to the cluster

ggilmore:


beyang:
Made some wording changes; otherwise LGTM

docs/install.md
```diff
@@ -30,6 +30,8 @@
    git clone https://github.com/sourcegraph/deploy-sourcegraph && cd deploy-sourcegraph && git checkout ${VERSION}
    ```
 
+1. If you want to change the number of `gitserver` replicas, follow [the instructions in configure.md](configure.md#configure-gitserver-replica-count) _before_ the next step.
```
nicksnyder: This might not be sufficient explanation. A person running through these steps might not know what gitserver is or why they would want more replicas.

Issue #84, opened 2018-09-07
Clarify recommended configurations


nicksnyder:
It is kind of expected that customers at least read through all the configurations listed in "common configurations". Maybe we should rename this to "recommended configurations"?

See discussion here:
https://github.com/sourcegraph/deploy-sourcegraph/commit/faacd647d13d72f3576155ea368e88cb0f8f8fa6#r30388357

cc @beyang

beyang:
People will not be guaranteed to read through anything unless it's placed on the main install path. There are a couple issues here:
* The "common configurations" list includes configurations that are not required: https://github.com/sourcegraph/deploy-sourcegraph/blob/master/docs/configure.md#common-configuration. That means that the storage class requirement will be interpreted as "optional".
* It is necessary to re-provision the volume when switching the storage class from the default to the one specifying SSDs. That means *immediately* after spinning up a Sourcegraph cluster, a user will have to blow away the original persistent volume (which may even require manual `kubectl` commands since the previous volume is already bound to the PVC).

If our intent (and I think it should be) is to have 99% of Data Center users use SSDs for their persistent volumes, we should just include it in the main install instructions here: https://github.com/sourcegraph/deploy-sourcegraph/blob/master/docs/install.md.

PR #87, opened 2018-09-11
add docs about the default storage class

ggilmore:
See https://github.com/sourcegraph/deploy-sourcegraph/issues/84


ggilmore:
I added a stub storage class (we can't specify a general one - the fields are hosting provider specific). 

`./kubectl-apply-all` will say that the storage class is invalid if you don't fill out the stub before applying the configuration (but the rest of the deployment will still succeed). Maybe we can create a storage class with a different name so that `kubectl` will really road-block you if you didn't specify the field?

base/default.StorageClass.yaml
```diff
@@ -0,0 +1,14 @@
+kind: StorageClass
```
nicksnyder: I don't think this should be included here because it is invalid. I think it would be sufficient to just inline this fragment into the README itself.

ggilmore:
@nicksnyder PTAL. I still need to verify that if you batch `kubectl apply` files, that it's guaranteed that the storage class is  config applied _before_ the PVCs are created.

docs/configure.md
```diff
@@ -377,7 +377,34 @@ See [the official documentation](https://kubernetes.io/docs/concepts/configurati
...
 ## Configure a storage class
 
-Sourcegraph relies on the default storage class of your cluster. If your cluster does not have a default storage class or if you wish to use a different storage class for Sourcegraph, then you need to update all PersistentVolumeClaims with the name of the desired storage class.
+**Note:** The storage class configuration steps should be done _before_ running `./kubectl-apply-all.sh` for the first time.
+
+By default, Sourcegraph uses the default storage class of your cluster. However, **we highly recommend that you use a storage class that uses SSDs as the underlying disk type**. The configuration details for the storage class differs depending on your hosting provider, so you should:
+
+1. Create a stub base/default.StorageClass.yaml.
+
+   ```yaml
```
nicksnyder: can we provide a concrete example for GCP at least, ideally AWS too?

docs/install.md
```diff
@@ -30,6 +30,8 @@
    git clone https://github.com/sourcegraph/deploy-sourcegraph && cd deploy-sourcegraph && git checkout ${VERSION}
    ```
 
+1. Configure the cluster's default storage class by reading through "Configure a storage class" in [docs/configure.md](../configure.md#configure-a-storage-class).
```
nicksnyder: make sure this gets applied first

docs/configure.md
ggilmore: Actually, this might not be strictly true. According to https://kubernetes.io/docs/concepts/storage/storage-classes/#gce, not specifying a `zone` will cause requests to round-robin between all the different zones that your nodes reside in (which might not be desirable).

nicksnyder: I still think an example helps.

For example you could have a commented out section that shows what a real GCP example would look like.

chrismwendt: The Azure storage class must be `managed-premium` https://github.com/sourcegraph/sourcegraph/issues/10795

ggilmore:
@nicksnyder PTAL

docs/configure.md
```diff
@@ -377,7 +377,52 @@ See [the official documentation](https://kubernetes.io/docs/concepts/configurati
 
 ## Configure a storage class
 
-Sourcegraph relies on the default storage class of your cluster. If your cluster does not have a default storage class or if you wish to use a different storage class for Sourcegraph, then you need to update all PersistentVolumeClaims with the name of the desired storage class.
+**Note:** The storage class configuration steps should be done _before_ running `./kubectl-apply-all.sh` for the first time.
```
nicksnyder: probably worth explaining what happens if you don't do this

ggilmore:
@nicksnyder PTAL

ggilmore:
@nicksnyder PTAL, there are more instructions that are needed so that the previously existing SC's are marked as non-default

docs/configure.md
```diff
@@ -377,7 +377,70 @@ See [the official documentation](https://kubernetes.io/docs/concepts/configurati
...
+    kubectl get sc -o json | jq --raw-output ".items | map(select(.metadata.name)) | .[] | \"kubectl patch sc -p '{\\\"metadata\\\":{\\\"annotations\\\":{\\\"storageclass.kubernetes.io/is-default-class\\\":\\\"false\\\"}}}' \\(.metadata.name)\"" | bash
+   ```
+
+1. In [create-new-cluster.sh](../create-new-cluster.sh), add the above command _before_ `./kubectl-apply-all`.
+
+   ```bash
+   #!/bin/bash
+
+   # Mark the existing storage classes in the cluster as non-default
+   kubectl get sc -o json | jq --raw-output ".items | map(select(.metadata.name)) | .[] | \"kubectl patch sc -p '{\\\"metadata\\\":{\\\"annotations\\\":{\\\"storageclass.kubernetes.io/is-default-class\\\":\\\"false\\\"}}}' \\(.metadata.name)\"" | bash
```
ggilmore: Actually, this won't work if you have a storage class with the `"storageclass.beta.kubernetes.io/is-default-class` annotation. I'm trying to figure out a jq script that can handle both cases.

nicksnyder:
If we are lifting the SSD requirement into the main install instructions, then maybe we should not mess with default storage class entirely.

We can pre-emptively use a custom storage class name in the base yaml (e.g. sourcegraph-ssd) which will require the creation of a non-default storage class with that name.

No only does this seem simpler, it more forcefully nudges customers to use ssds.

docs/install.md
```diff
@@ -30,6 +30,8 @@
    git clone https://github.com/sourcegraph/deploy-sourcegraph && cd deploy-sourcegraph && git checkout ${VERSION}
    ```
 
+1. Configure the `sourcegraph` storage class for the cluster by reading through "Configure a storage class" in [docs/configure.md](../configure.md#configure-a-storage-class).
```
beyang: `["Configure a storage class"](../configure.md#configure-a-storage-class).`

The link text should correspond to what's linked to.

ggilmore:
@nicksnyder @beyang PTAL, I added migration instructions concerning the storage classes and added snippets for the different hosting providers for using SSDs (I'm not sure if the value are correct for AWS and Azure - this is only from looking at the documentation). 

@beyang - @nicksnyder and I talked about this, but configuring SSDs this way introduces a breaking change since the storage class name is newly specified.

PR #95, opened 2018-09-16
include `provisioner` field in docs

beyang:
When I was going through these install instructions, I missed the `provisioner` field the first time because I was just copy-pasting the snippets. Any reason not to include these?


ggilmore:
No, but the examples aren't necessarily copy-pastable on their own since the rest of the parameters depends on your hosting provider / cluster configuration

PR #102, opened 2018-10-12
xlang: add instruction in docs to add entry to site configuration when you enable a language server

ggilmore:
This drops the `LANGSERVER_*` environment variables (except the ones ending in `_BG`) in favor of adding entries to the `langservers` array in site config because the Sourcegraph extension corresponding to a language server will not be enabled unless there is an entry for it in the `langservers` array in site config.

The `LANGSERVER_*` environment variables will enable the language server, but not the corresponding Sourcegraph extension.

cc @chrismwendt

### DIFF
--- a/configure/xlang/experimental/bash/README.md
+++ b/configure/xlang/experimental/bash/README.md
@@ -12,7 +12,7 @@ You can enable it by:
    echo kubectl apply --prune -l deploy=xlang-bash -f configure/experimental/bash --recursive >> kubectl-apply-all.sh
    ```
 
-2. Add the following environment variables to the `lsp-proxy` deployment to make it aware of the Bash language server's existence.
+1. Add the following environment variables to the `lsp-proxy` deployment to make it aware of the Bash language server's existence.
 
    ```yaml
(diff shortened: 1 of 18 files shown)


nicksnyder:
Please update PR description to explain why we are doing this (I thought env variable was sufficient).

ggilmore:
It isn't anymore due to changes w/ Sourcegraph extensions. @chrismwendt can talk about the relevant changes

nicksnyder:
either way, it belongs in the PR description so it ends up in the commit message

chrismwendt:
I spun up a data center instance with `LANGSERVER_GO` and without `langservers`, and code intelligence worked, but the site-admin/code-intelligence page did not show that the language server was enabled.

`LANGSERVER_*` env vars (except the `_BG` ones) are no longer recommended, so I removed those instructions from these docs and replaced them with instructions to add `"address": "tcp://xlang-<LANG>:<PORT>"` to the `langservers` array entry.

ggilmore:
cc @nicksnyder, updated

nicksnyder:
related: https://github.com/sourcegraph/enterprise/issues/11871

Issue #127, opened 2018-10-22
Publish seperate zoekt 'indexserver' and 'webserver' images with more defaults baked in


slimsag:
Today, users must specify a ton of arguments to our `sourcegraph/zoekt` image to run the indexserver and webserver:

https://github.com/sourcegraph/deploy-sourcegraph/blob/bd53e169b5c0405ad4db1114c780cb5cdc729a2c/base/indexed-search/indexed-search.Deployment.yaml#L55-L68

https://github.com/sourcegraph/deploy-sourcegraph/blob/bd53e169b5c0405ad4db1114c780cb5cdc729a2c/base/indexed-search/indexed-search.Deployment.yaml#L27-L34

It would be much better from a pure-Docker deployment perspective to publish separate `sourcegraph/zoekt-indexserver` and `sourcegraph/zoekt-webserver` Docker images with many of these defaults baked in. i.e. only `-sourcegraph_url` should be required for indexserver, and nothing required for webserver.

keegancsmith:
Note that both of these images need to share a volume (`/data/index`). We can bake in these defaults and make it different images. I'm interested to hear your motivation for that though.

slimsag:
Oh, nice, I didn't know that. Thanks! That saves me from tracking down a bug :)

My motivation is simple: I'm trying to make a [pure-Docker cluster deployment option](https://github.com/sourcegraph/deploy-sourcegraph-docker) and any extra args etc. that we need to specify make our deployment config that much more complex.

Issue #124, opened 2018-10-22
redis-cache/redis-store: Bake configuration and command into image


slimsag:
We publish our own Redis image, so we should be baking the configuration + command to run directly into the image:

https://github.com/sourcegraph/deploy-sourcegraph/blob/bd53e169b5c0405ad4db1114c780cb5cdc729a2c/base/redis/redis-cache.Deployment.yaml#L28-L30

This will significantly help simplify a pure-Docker deployment.

Issue #122, opened 2018-10-22
gitserver: SRC_REPOS_DIR should default to /data/repos


slimsag:
i.e. we should not need to set this:

https://github.com/sourcegraph/deploy-sourcegraph/blob/bd53e169b5c0405ad4db1114c780cb5cdc729a2c/base/gitserver/gitserver.StatefulSet.yaml#L29-L30

Issue #121, opened 2018-10-22
github-proxy: LOG_REQUESTS should default to 'true'


slimsag:
i.e. we should not need to set this:

https://github.com/sourcegraph/deploy-sourcegraph/blob/bd53e169b5c0405ad4db1114c780cb5cdc729a2c/base/github-proxy/github-proxy.Deployment.yaml#L30-L31

PR #134, opened 2018-11-02
major simplifications

### LINKED ISSUE #123: pgsql: rely on defaults for PG vars
We deploy our own image for Postgres, so we should just bake these environment variables into the images by default:

https://github.com/sourcegraph/deploy-sourcegraph/blob/bd53e169b5c0405ad4db1114c780cb5cdc729a2c/base/pgsql/pgsql.Deployment.yaml#L29-L33

This helps simplify a pure-Docker deployment.

slimsag:
This updates images to new versions that contain many more defaults. See individual commit messages for details.

I did not test this on a real cluster, but I ran the YAML through `kubeval` (validator) and the same changes applied to deploy-sourcegraph-docker work fine.

Fixes #123 #127 #124 #132 #118 #119 #121 #120 #122


docs/configure.md
```diff
@@ -312,7 +312,7 @@ Increasing the number of `gitserver` replicas can improve performance when your
 To change the number of `gitserver` replicas:
 
 1. Update the `replicas` field in [gitserver.StatefulSet.yaml](../base/gitserver/gitserver.StatefulSet.yaml).
-1. Update the `SRC_GIT_SERVERS` environment variable in all services to reflect the number of replicas.
+1. Update the `SRC_GIT_SERVERS` environment variable in the frontend service to reflect the number of replicas.
```
nicksnyder: Please update the convenience script below (L353)

nicksnyder:
I think this is worth testing on a real cluster before merging.

ggilmore:
I'd also like to see this tested before merging.

docs/configure.md
slimsag: I looked into this just now.

(1) It does still work, but is more complex than we need in the long run.
(2) We cannot simplify it right now because the xlang-go and xlang-go-bg deployments still need to have gitservers updated (but this is changing soon as those are removed). We can simplify it once those are removed.

slimsag:
Tested locally, found and fixed a bug, and confirmed working now.

PR #117, opened 2018-10-22
DO NOT MERGE: Remove all references to lsp-proxy and xlang-go's gitserver dependency 

ggilmore:
This will be the impact of the things that you are working on. You should test the impact of your changes by trying to deploy this branch and seeing if your instance still works.

PR #140, opened 2018-11-08
WIP (DON'T MERGE) - add stub files for management console

ggilmore:


base/config-file.ConfigMap.yaml
```diff
@@ -1,20 +0,0 @@
-apiVersion: v1
-data:
-  config.json: |-
```
ggilmore: @slimsag I think we'll need to make this the default value by other means

ggilmore:
@slimsag We shouldn't need do delete references to `sg-config` in `configure/xlang/...` since https://github.com/sourcegraph/deploy-sourcegraph/pull/117 should be handling it

PR #146, opened 2018-11-26
rename appURL -> externalURL

ggilmore:
Fixes https://github.com/sourcegraph/sourcegraph/issues/933

This PR shouldn't be merged until the `3.0-preview` images have been tagged.

### DIFF
--- a/base/config-file.ConfigMap.yaml
+++ b/base/config-file.ConfigMap.yaml
@@ -3,7 +3,7 @@ data:
   config.json: |-
     {
       // Publicly accessible URL to web app (e.g., what you type into your browser).
-      "appURL": "http://localhost:3080",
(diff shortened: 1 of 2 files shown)


slimsag:
@ggilmore please also update https://github.com/sourcegraph/deploy-sourcegraph-docker

slimsag:
i.e. just https://github.com/sourcegraph/deploy-sourcegraph-docker/blob/master/sourcegraph/config.json

slimsag:
I am merging this now because:

(1) #134 bumped us to an alpha version which should contain this change I think.
(2) If (1) above is not true, we can address it after merging. i.e., users are using the tagged versions of this repository.
(3) This change is needed for me to merge the management console (#151) because this config file will still be used for 3.0 for existing user migration purposes.

PR #147, opened 2018-11-26
Link to helpful instructions for using Docker for Mac's Kubernetes support for testing

slimsag:
I wrote these instructions for myself for testing our YAML here on a Docker for Mac Kubernetes cluster. I am not sure if these are beneficial for others on the team or not, so I kept the section very small and linked to a separate gist to really drive home that these are more personal notes.

In fact, I'm not sure if merging this is worth it. Thoughts? At the very least, I am opening this PR so it is findable via GitHub PR search.

### DIFF
--- a/README.dev.md
+++ b/README.dev.md
@@ -20,7 +20,7 @@ Wait for buildkite to pass and for your changes to be approved, then merge and c
 
 Test what is currently checked in to master by [installing](docs/install.md) Sourcegraph on a fresh cluster.
 
-### Provision a new cluster
+### Option 1: Provision a new cluster on GCP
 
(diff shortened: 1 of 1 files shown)


slimsag:
cc @ggilmore @nicksnyder if either of you think this is a good addition let me know.

nicksnyder:
I think the Docker for Mac instructions are worth checking in somewhere. Perhaps in their own file at first (not as part of the test/release process instructions).

If we could make it easy/reliable/scriptable to setup/test a cluster locally in the future, I could see that being the preferred way to test releases (I think I would be more likely to test if I could do it entirely locally).

slimsag:
Yeah, makes sense.

I agree about the easy part, but I couldn't find a way to bypass Kubernetes resource limits so you basically end up hand-editing several files which is iffy

nicksnyder:
I think the editing is something that could be scripted

slimsag:
yeah, that may make sense if there is not a better alternative somewhere

PR #153, opened 2018-12-14
Fix link to configuring git server

zoidbergwill:


ggilmore:
Thanks @zoidbergwill !

PR #156, opened 2019-01-08
Exclude raw route from page load latency alert

keegancsmith:
It isn't expected to perform with the same latency as interactive endpoints.


ggilmore:
Lgtm

PR #160, opened 2019-01-11
ci: remove kubeval

ggilmore:
Given that we already validate the yaml by running `kubectl apply -f ... --validate --dry-run`, running `kubeval` seems superflous. 

Also, `kubeval` seems to suffer from both [false positives](https://github.com/garethr/kubeval/issues/93) and [false negatives (responsible for this build failure)](https://buildkite.com/sourcegraph/deploy-sourcegraph-dot-com/builds/126#29383e20-8d0e-4ff8-ab24-6074cc3093f2).


ggilmore:
> If we can reproduce false positives with kubeval it would be nice to file a bug report on that project.

You already did in https://github.com/garethr/kubeval/issues/93

ggilmore:
Whoops yeah, I had the definitions flipped in my head for what `kubeval` is doing

PR #161, opened 2019-01-14
pgsql: Upgrade to Postgres 11.1

tsenart:
This commits upgrades our Kubernetes deployment of Postgres to use 11.1
in time for the 3.0-beta release.

I've created placeholder documentation of the upgrade procedure for existing
deployments which I'll work on this week.


tsenart:
@keegancsmith: Hmm, the goal with this PR was that new deployments of Sourcegraph can use Postgres 11.1. Is there a way for us to parameterise the version that is used in production and still allow new users to default to Postgres 11.1?

ggilmore:
You'd have to communicate to the team that merging in upstream for https://github.com/sourcegraph/deploy-sourcegraph-dot-com is unsafe until you figure that out and either:

1) Hold off merging upstream
2) Communicate which parts of the diff should be ignored when merging in upstream

tsenart:
@ggilmore: My question was whether it would be possible to parameterise the Postgres version in the Kubernetes manifests, ensure production is set with 9.4, and make the default 11.1 for new users.

keegancsmith:
You would have to merge this in, then when you merge it into the release branch for our production fork ensure it points to the current version.

PR #178, opened 2019-01-26
Add instructions for nginx without ingress controllers

beyang:


base/config-file.ConfigMap.yaml
```diff
@@ -17,7 +17,9 @@ data:
       //
       // Other providers are documented here:
       // https://about.sourcegraph.com/docs/config/site#authproviders-array
-      "auth.providers": [{"type": "builtin", "allowSignup": false}]
+      "auth.providers": [{"type": "builtin", "allowSignup": false}],
+
+      "licenseKey": "eyJzaWciOnsiRm9ybWF0Ijoic3NoLXJzYSIsIkJsb2IiOiJKWlRJam8xZzV3TlBrcDE3VUlGekVnN1o5a25BcUR1aUNsb1FpbHNXbU1GV0ZkZ2tueUhCMVNWYmJDWFdrT3FZaDB2SHk0bWFGYXd4VkpDdTMyNTU3cW96SmhUd1hLcWlYTU9BRzFvb0s3Y09NTlI4SEtNbmYzS2RFUkI0czR4YzM3RmRnODRIZWlBdHNHeWwwejVHNkFjOVB1RmNQM2kvUWQyeWdjdTdpdUFQRnJtNVpQQWJSa0dXaktmL1F4WDVScENFOHpWd2dKZmViclphSmZWUnozRHR0eUdBYThhczYzMlgzZk1NNlJ6c2F0NS8yT3BHRFFhbWlXNWdTU1hpQ2JycDB3SlNkbXk4T3hpSTU0NVN0YVpmbWk5ZWNtNERoS1Z6TDJyWXRid2M5b3A0YkNhbWJXb0VJTGI1UGFVdUdkVjF2YVpseEREVlN4Q3huYlJNSnc9PSJ9LCJpbmZvIjoiZXlKMklqb3hMQ0p1SWpwYk5Ua3NPU3d6Tnl3eE5pd3hPRFVzTXpJc01UWXhMREl4T0Ywc0luUWlPbHNpZEhKMVpTMTFjQ0pkTENKMUlqb3hMQ0psSWpvaU1qQXhPUzB4TUMwd05WUXhPVG93TlRvd09Gb2lmUT09In0"
```
keegancsmith: @beyang did you mean to add this licenseKey into this repo?

base/frontend/sourcegraph-frontend.Service.yaml
```diff
@@ -11,9 +11,8 @@ metadata:
 spec:
   ports:
   - name: http
-    nodePort: 30080
     port: 30080
     targetPort: http
   selector:
     app: sourcegraph-frontend
-  type: NodePort
+  type: ClusterIP
```
keegancsmith: If people try apply this, won't it fail if the service is already exists (as a NodePort)?

keegancsmith: I agree, it was a change I wanted to make but did not due to this issue. If so we need to document it in the migration guide. An alternative solution is to update the selector on this service so it matches nothing, and create a new service. I'm unsure which is preferred. Likely the former since an admin upgrading would likely want to be aware of this sort of change (and will need to be doing changes for ingress-nginx anyways).

nicksnyder:
What is the decision here? In the absence of a decision or merging this, whatever is in master is going to get tagged as 3.0.0.

beyang:
@ggilmore PTAL (addressed your comments from https://sourcegraph.slack.com/archives/D1B7C7WQN/p1548724204000200).

Re the config map: changes to the config map will not propagate to nginx without a restart, but that should be known to anyone who's running nginx (i.e., I think discussing this is beyond the scopes of what our docs should be)

beyang:
I'll own deploying to sourcegraph.com

beyang:
Merged manually in 5a3afdfa1968c610b9a121ee3ce9eeb08232007c

PR #179, opened 2019-01-29
pgsql: Use smarter readiness and liveness probes

keegancsmith:
Our pgsql image automatically upgrades from 9.4 to 11.1. However, this can be a
slow process leading to the liveness probe killing the pod before the migration
is finished. Our image now embeds scripts to check the liveness and readiness:

- readiness checks if postgres is available via TCP us pg_isready
- liveness passes if we are upgrading. Otherwise it uses the readiness check.

This should also fix the postgres log spam about bad TCP clients, since we are
removing the dumb TCP probe we used to have.

Part of https://github.com/sourcegraph/sourcegraph/issues/1404

### DIFF
--- a/base/pgsql/pgsql.Deployment.yaml
+++ b/base/pgsql/pgsql.Deployment.yaml
@@ -26,7 +26,7 @@ spec:
     spec:
       containers:
       - env:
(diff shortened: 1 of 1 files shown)


tsenart:
I haven't seen a PR for those image changes. Was there one?

tsenart:
Looks good to me.

PR #180, opened 2019-02-01
Revert Postgres to 9.4 for the 3.0 release

tsenart:
This commit reverts back from the 11.1 Postgres version since we couldn't finish everything in time.

Follow up: https://github.com/sourcegraph/sourcegraph/issues/1404#issuecomment-459900259

### DIFF
--- a/base/pgsql/pgsql.Deployment.yaml
+++ b/base/pgsql/pgsql.Deployment.yaml
@@ -26,7 +26,7 @@ spec:
     spec:
       containers:
       - env:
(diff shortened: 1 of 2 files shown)


nicksnyder:
Why isn't this a strict revert of https://github.com/sourcegraph/deploy-sourcegraph/pull/161/files?

nicksnyder:
Ah, I see https://github.com/sourcegraph/deploy-sourcegraph/pull/179

tsenart:
@nicksnyder: That and also: https://github.com/sourcegraph/infrastructure/pull/1125

PR #196, opened 2019-02-18
Delay frontend liveness check to 5 minutes

nicksnyder:
The frontend requires other services to be running, so it needlessly enters CrashLoopBackoff during normal startup.

### DIFF
--- a/base/frontend/sourcegraph-frontend.Deployment.yaml
+++ b/base/frontend/sourcegraph-frontend.Deployment.yaml
@@ -63,7 +63,7 @@ spec:
             path: /healthz
             port: http
             scheme: HTTP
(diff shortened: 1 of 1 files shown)


nicksnyder:
The current is 60 seconds and it restarts twice so that is ~180 seconds. I didn't wan't to fiddle with it any more so 5 minutes is safe.

If there is a config issue restarting the instance won't help anyway.

Liveness checks should be inherently conservative on startup.

nicksnyder:
I may have incorrectly assumed that these restarts were caused by a 1 minute liveness check instead of application logic that panics after 30 seconds, as discussed here: https://github.com/sourcegraph/sourcegraph/issues/2812#issuecomment-474119891

PR #197, opened 2019-02-19
Use semver versionScheme for sourcegraph images

rarkins:
This tells Renovate to assume semver-compatible versioning for all sourcegraph images. By default Renovate assumes typical Dockerish versions, where the suffix is related to platform (e.g. `3.1.0-alpine` and not semver-like (e.g. `3.1.0-rc.3`).


nicksnyder:
Awesome, thanks! I was just about to try this out :)

PR #202, opened 2019-02-21
add category:rbac label

ggilmore:
Deploying [RBAC](https://kubernetes.io/docs/reference/access-authn-authz/rbac/) manifests to a Kubernetes cluster might require elevated permissions (which could be undesirable in some CD scenarios).

This PR adds a `category: rbac` label to all RBAC manifests. This makes it possible to possible [to exclude these manifests](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) from certain `kubectl` operations (e.g `-l deploy=sourcegraph,category!=rbac)


nicksnyder:
Great idea!

PR #217, opened 2019-04-15
Do not specify PG* envvars in repo-updater

keegancsmith:
Since https://github.com/sourcegraph/sourcegraph/pull/3346 repo-updater relies on the ServiceConnections config it fetches from the frontend. Now admins do not need to configure multiple deployments when updating PG settings.

Part of https://github.com/sourcegraph/sourcegraph/issues/2025

PR #218, opened 2019-04-15
Do not specify PG* envvars in repo-updater

keegancsmith:
Since https://github.com/sourcegraph/sourcegraph/pull/3346 repo-updater relies on the ServiceConnections config it fetches from the frontend. Now admins do not need to configure multiple deployments when updating PG settings.

Redo of https://github.com/sourcegraph/deploy-sourcegraph/pull/217

Part of https://github.com/sourcegraph/sourcegraph/issues/2025

### DIFF
--- a/base/repo-updater/repo-updater.Deployment.yaml
+++ b/base/repo-updater/repo-updater.Deployment.yaml
@@ -27,16 +27,6 @@ spec:
       containers:
       - image: sourcegraph/repo-updater:3.2.2@sha256:5b8a3d31405e0cd5f7343a0ee6453fc31ad50e0e511d5147a836db9989daae94
         env:
-        - name: PGHOST
-          value: pgsql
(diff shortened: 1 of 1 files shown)


keegancsmith:
@slimsag when you update this repo to have 3.3 as part of the release, can you ensure this PR is merged? We can't merge it until 3.3 and ideally it would be part of that release. In future we should probably have a next branch / etc.

slimsag:
I will do it, but I don't understand why you can't just merge it now? We tag releases and `master` is not considered stable for customers, they should always be using a tag or release branch.

keegancsmith:
I can just merge. But then the state of master is broken since the image used currently requires the envvars. @ggilmore is that acceptable?

slimsag:
Ah I get it now. In that case I'm not sure, I'll merge before release
though.

PR #230, opened 2019-05-02
add back language extensions to deploy-sourcegraph

ggilmore:
TODO: 

- [x] Add example TLS configurations for Go, Typescript
   - [x] ask @chrismwendt if it's safe to not use https if they're not exposed over the internet
   - [x] add example Ingress configurations
   - [ ] ~add example TLS_CERT, TLS_KEY configurations~ -> not handling non- nginx-ingress deployments for now
   - [x] write docs to show how to use http basic auth with nginx-ingress

- [x] write docs for configure.md
- [x] figure out whether or not resource limits/requests are reasonable 
- [ ] switch docker tags to non-`insiders` ones

---

Test plan: 

1. Spin up a new test cluster with https://github.com/sourcegraph/deploy-k8s-helper/tree/lang-test and `deploySourcegraphRoot` checked out to https://github.com/sourcegraph/deploy-sourcegraph/tree/add-back-lang-ext

1. Use mkcert to generate a certificate for sourcegraph.example.com

   ```console
   > mkcert  -cert-file cert -key-file key 35.232.68.120
   ```

1. Run `cat ./cert | pulumi config --secret set certificate` and `cat ./cert | pulumi config --secret set key`

1. Create htpassword using instructions in https://github.com/sourcegraph/deploy-sourcegraph/blob/add-back-lang-ext/configure/lang/go/README.md

    ```console
    > htpasswd -c auth langserveruser 
    ```

1.  Run `cat ./auth | pulumi config --secret set htpassword`

1. Run `yarn up`

1. Add the ip address from ingressIPs to `/etc/hosts`

1. Add an external service sourcegraph.example.com, clone sourcegraph/sourcegraph on it 

1. Use Global settings from https://github.com/sourcegraph/deploy-sourcegraph/blob/add-back-lang-ext/configure/lang/go/README.md and https://github.com/sourcegraph/deploy-sourcegraph/blob/add-back-lang-ext/configure/lang/typescript/README.md to connect to the language servers

1. Open a Go and Typescript file from sourcegraph/sourcegraph. You should see code intelligence, and your network tab should show succesfull communication to the language servers


configure/lang/go/lang-go.Deployment.yaml
```diff
@@ -0,0 +1,72 @@
...
+        - -usebinarypkgcache=false
+        - -cachedir=$(CACHE_DIR)
+        - -freeosmemory=false
+        env:
+        - name: POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+        - name: CACHE_DIR
+          value: /mnt/cache/$(POD_NAME)
```
chrismwendt: TLS env vars from https://github.com/sourcegraph/deploy-sourcegraph-dot-com/blob/8f7a33406c6d2114abaa55e99a1b4cd849956dfe/configure/lang/go/lang-go.Deployment.yaml#L46-L55

```yaml
        - name: TLS_CERT
          valueFrom:
            secretKeyRef:
              key: cert
              name: tls
        - name: TLS_KEY
          valueFrom:
            secretKeyRef:
              key: key
              name: tls
```

configure/lang/go/lang-go.Deployment.yaml
```diff
@@ -0,0 +1,72 @@
...
+            port: 7777
+        terminationMessagePolicy: FallbackToLogsOnError
+        resources:
+          limits:
+            # TODO@ggilmore: Are these figures reasonable?
+            cpu: "4"
+            memory: 8Gi
+          requests:
+            cpu: "1"
+            memory: 2Gi
```
chrismwendt: These look reasonable to me compared to production https://github.com/sourcegraph/deploy-sourcegraph-dot-com/blob/8f7a33406c6d2114abaa55e99a1b4cd849956dfe/configure/lang/go/lang-go.Deployment.yaml#L79-L84

configure/lang/go/lang-go.Deployment.yaml
ggilmore: @chrismwendt I had to remove them from the default deployment since not everyone will have a secret named `tls` in their system. We should document the existence of these env vars, but they can't be part of the default deployment.

ggilmore:
@chrismwendt @beyang @slimsag PTAL

 I still need to test the non-pulumi instructions, but I think this is in a good enough state for review.

ggilmore:
cc @keegancsmith since I am editing the ingress manifests that you added

base/frontend/sourcegraph-frontend.Ingress.yaml
```diff
@@ -17,15 +17,17 @@ spec:
...
+  - host: sourcegraph.example.com
+    http:
       paths:
       - path: /
         backend:
           serviceName: sourcegraph-frontend
           servicePort: 30080
     # Set a host if you have multiple ingresses. This installs a default
     # handler.
+    # TODO@ggilmore: Is it safe to switch the host?
```
keegancsmith: what do you mean by switch the host?

ggilmore: @keegancsmith I meant using an explicit host instead of relying on the default handler

configure/lang/go/README.md
```diff
@@ -0,0 +1,76 @@
...
+TLS/SSL is required for secure communication with the language server. Once you have completed ["Configure TLS/SSL"](../../../docs/configure.md#configure-tlsssl) in [docs/configure.md](../../../docs/configure.md#configure-tlsssl), change the `sourcegraph.example.com` domain in [configure/lang/go/lang-go.Ingress.yaml](lang-go.Ingress.yaml) to the real domain name that you are using for your Sourcegraph instance.
+
+```diff
+spec:
+    tls:
+    - hosts:
+    # 🚨 TLS is required for secure communication with the language server. 
+    # See the customization guide (../../../docs/configure.md) for information
+    # about configuring TLS
+-     - sourcegraph.example.com
```
beyang: Why not keep "sourcegraph.example.com"? That way someone can replace all instances of it at the same time?

ggilmore:
@beyang I've removed references to `your.real.domain.com` in the langserver READMEs, and just added a yaml comment instead. Could you see if this address your comment?

I'll test the manual instructions now, but I would appreciate another pair of eyes.

ggilmore:
@keegancsmith @chrismwendt @beyang @slimsag PTAL

After some debugging with @slimsag yesterday, we discovered using that using HTTP Basic auth for the language server doesn't work if the language server doesn't have its own domain. 

The HTTP Basic auth spec says that it's safe for browsers to assume that all requests to the same host use the same basic auth credentials. However, only the requests to the language server use basic auth, _not_ to the frontend api. This means that all requests to the API will fail with `401 unauthorized `(because the browser is re-using the same credentials, which the spec says is safe to do). This completely breaks the web app in Firefox. 

I've updated the instructions to say that you _must_ put the language servers on their own domain. 

I suspect that this behavior is broken in the single docker image case as well. I'll need to follow up afterward to confirm.

configure/lang/go/README.md
```diff
@@ -0,0 +1,96 @@
...
+        rules:
+        - host: go.sourcegraph.example.com
+    ```
+
+**WARNING:** Do NOT commit the actual TLS cert and key files to your fork (unless your fork is
+private **and** you are okay with storing secrets in it).
+
+### HTTP basic authentication
+
+HTTP basic authentication is used to prevent unauthorized access to the language server. At a high level, you'll create a secret then put it in both [configure/lang/go/lang-go.Ingress.yaml](lang-go.Ingress.yaml) and in your Sourcegraph global settings so that logged-in users are authenticated when their browser makes requests to the Go language server.
```
slimsag: This does mean that if that basic auth credential is exposed by any Sourcegraph user, anyone with that credential now has unauthorized access to the language server and the code it operates on. Shouldn't we warn people about that? i.e. basic auth is better than nothing, but really you need an actual auth proxy secure your langservers.

configure/lang/go/lang-go.Deployment.yaml
slimsag: I am surprised we are setting these so low. I suspect any large repositories like Kubernetes will OOM with this few resources

configure/lang/go/README.md
ggilmore: @slimsag 
This is possible, but this security model is the same as what we currently have as a result of https://github.com/sourcegraph/sourcegraph/issues/2595. We should have a discussion about whether or not this is sufficient, but that shouldn't block this PR.

ggilmore: Note that this PR means that part of https://github.com/sourcegraph/sourcegraph/issues/2595#issuecomment-470014851 is now invalid

slimsag: Filed https://github.com/sourcegraph/security-issues/issues/16

configure/lang/go/lang-go.Deployment.yaml
ggilmore: @slimsag  Do you think it's better to bump this to 12 gigs of memory, etc. even if not every user will need it?

configure/lang/go/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+**WARNING:** 🚨 If your basic auth credentials are exposed, anyone with that credential now has unauthorized access to the language server and the code it operates on. Using an auth proxy, VPN, or firewall would provide more secuirty. More information about this alternative methods will come at a later date. 🚨
```
slimsag:
```suggestion
**WARNING:** 🚨 If your basic auth credentials are exposed, anyone with that credential now has unauthorized access to the language server and the code it operates on. Using an auth proxy, VPN, or firewall would provide more security. More information about this alternative methods will come at a later date. 🚨 
```

configure/lang/typescript/README.md
```diff
@@ -0,0 +1,98 @@
...
+    ```
+
+**WARNING:** Do NOT commit the actual TLS cert and key files to your fork (unless your fork is
+private **and** you are okay with storing secrets in it).
+
+### HTTP basic authentication
+
+HTTP basic authentication is used to prevent unauthorized access to the language server. At a high level, you'll create a secret then put it in both [configure/lang/typescript/lang-typescript.Ingress.yaml](lang-typescript.Ingress.yaml) and in your Sourcegraph global settings so that logged-in users are authenticated when their browser makes requests to the Javascript/Typescript language server.
+
+**WARNING:** 🚨 If your basic auth credentials are exposed, anyone with that credential now has unauthorized access to the language server and the code it operates on. Using an auth proxy, VPN, or firewall would provide more secuirty. More information about this alternative methods will come at a later date. 🚨
```
slimsag:
```suggestion
**WARNING:** 🚨 If your basic auth credentials are exposed, anyone with that credential now has unauthorized access to the language server and the code it operates on. Using an auth proxy, VPN, or firewall would provide more security. More information about this alternative methods will come at a later date. 🚨
```

configure/lang/go/lang-go.Deployment.yaml
slimsag: I think there is a high likelihood that users deploying a Kubernetes cluster AND Go language server will also have at least one very large repository. Basically I think the default is incorrectly inverted currently.

I would set to the values prod uses, then add a comment that you may lower them to the values currently here if you do not have large Go repositories.

PR #244, opened 2019-05-22
Update documentation for Go language server deployment

mrnugget:
This commit changes the documentation for the Go language server
deployment to make it clearer that, you can try out the Go language
server support...

1. ... without having to setup TLS/SSL (even though we recommend it!)
2. ... without having to use HTTP basic authentication (even though we recommend it!)

In order to that, I did three things:

1. add a suffix to the two sections so it's clear you can skip the "Setup
   TLS/SSL" and "HTTP Basic authentication" section
2. extract the updating of the global settings into a separate step so
   readers won't miss it when they skip HTTP basic auth
3. extend the section about global settings to explain what to do when
   the first two sections were skipped

### DIFF
--- a/configure/lang/go/README.md
+++ b/configure/lang/go/README.md
@@ -4,11 +4,11 @@ This folder contains the deployment manifests for the [Go language extension](ht
 
 ## Installation instructions
 
-### Setup TLS/SSL 
+### Setup TLS/SSL (Highly recommended, optional)
 
 TLS/SSL is required for secure communication with the language server. Once you have completed ["Configure TLS/SSL"](../../../docs/configure.md#configure-tlsssl) in [docs/configure.md](../../../docs/configure.md#configure-tlsssl), for your overall Sourcegraph instance, you'll need to configure TLS/SSL for the Go language server as well.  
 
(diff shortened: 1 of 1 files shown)


mrnugget:
Happy to change any of the changes (I can totally understand if we never want to say that TLS is optional, even though I guess that the audience of this document knows that you always _upgrade_ to TLS). I just wanted to update the docs to make it clear that it's actually pretty easy to deploy a Go language server with these instructions - it took me longer to find out what of the docs applies to my case and what doesn't than to actually deploy it :)

keegancsmith:
I think @chrismwendt is the strongest owner here, so will defer to him for accepting it.

mrnugget:
> What has been suggested is that this doc be split into different deployment types: k8s, single-Docker image, and local testing. I think that restructuring should happen before we add more deployment types.

Ah, I see, that makes sense. But just so we're on the same page: I was under the assumption that _this_ document here is only about k8s. It doesn't mention single Docker images or local testing. What I'm doing in this PR is adding a little "k8s-testing" information.

Are you maybe thinking of the `sourcegraph-go` README? https://github.com/sourcegraph/sourcegraph-go Because that one contains Docker, k8s, etc. in one document

PR #258, opened 2019-06-10
update prometheus cluster role so it can provide node metrics

beyang:
A customer applied this patch to their Prometheus cluster role so they could use https://github.com/braedon/prometheus-es-exporter

### DIFF
--- a/configure/prometheus/prometheus.ClusterRole.yaml
+++ b/configure/prometheus/prometheus.ClusterRole.yaml
@@ -9,10 +9,13 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - nodes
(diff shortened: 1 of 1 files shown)


keegancsmith:
Why did the customer have to patch this role? Shouldn't they just create a new role for the exporter? I think we should be hesitant to expand the default set of permissions prom needs, especially since prometheus doesn't actually need them (but this exporter did)

beyang:
@keegancsmith just tagged you on the customer issue

keegancsmith:
aah I this now makes a lot of sense. I was wondering why we lost our cAdvisor metrics at one point. SGTM

Issue #263, opened 2019-06-18
Images should use FQDN (docker.io/foo/bar not foo/bar)


slimsag:
It is possible to configured your Kubernetes cluster to default to a registry other than docker.io and this appears to be somewhat common, so we should be using FQDN everywhere that we specify images.

PR #271, opened 2019-06-20
use fdqn for sourcegraph docker hub images

ggilmore:
Fixes https://github.com/sourcegraph/deploy-sourcegraph/issues/263


ggilmore:
cc @sourcegraph/distribution

PR #277, opened 2019-07-03
base/syntect-server: increase memory from 512 MB -> 4 GB

slimsag:
Various deployments we have show that with some languages syntect_server
consumes more memory and OOMs. Bumping this up to 4 GB makes sense and
has worked well in customer deployments that previously had issues.


slimsag:
I am merging this without review / approval because I am very confident in this change being correct and given it is based on existing data / customer deployments, there is not much here for anyone to review anyway.

slimsag:
Cherry-picking this onto 3.5.1

PR #284, opened 2019-07-15
Add lsif-server

chrismwendt:
This adds lsif-server that was introduced in https://github.com/sourcegraph/sourcegraph/pull/4799

lsif-server is a service that sits behind Sourcegraph's frontend and exposes an HTTP API that receives file uploads and stores them on disk. That storage is expected to stick around for the lifetime of the Sourcegraph instance.

TODO

- [x] Add the 3.6 image tag
- [x] Set up a persistent volume (@sourcegraph/distribution could anyone help me with this?)

### DIFF
--- a/base/lsif-server/lsif-server.Deployment.yaml
+++ b/base/lsif-server/lsif-server.Deployment.yaml
@@ -65,4 +65,5 @@ spec:
         runAsUser: 0
       volumes:
       - name: lsif-storage
-        emptyDir: {}
+        persistentVolumeClaim:
+          claimName: lsif-server
--- a/base/lsif-server/lsif-server.Deployment.yaml
+++ b/base/lsif-server/lsif-server.Deployment.yaml
@@ -1,0 +1,68 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    description: LSIF HTTP server for code intelligence.
+  labels:
+    deploy: sourcegraph
+  name: lsif-server
+spec:
+  minReadySeconds: 10
+  replicas: 1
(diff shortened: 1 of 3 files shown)


base/lsif-server/lsif-server.Deployment.yaml
```diff
@@ -0,0 +1,68 @@
...
+            cpu: 500m
+            memory: 500M
+        volumeMounts:
+        - mountPath: /lsif-storage
+          name: lsif-storage
+      securityContext:
+        runAsUser: 0
+      volumes:
+      - name: lsif-storage
+        emptyDir: {}
```
chrismwendt: I set up a volume mount, but I don't think this is persistent. It looks like psql has what I'm looking for, so could I copy it?

https://github.com/sourcegraph/deploy-sourcegraph/blob/48c69935ec30f702ea998e8bad431c8f7a74dd53/base/pgsql/pgsql.Deployment.yaml#L69-L72

https://github.com/sourcegraph/deploy-sourcegraph/blob/48c69935ec30f702ea998e8bad431c8f7a74dd53/base/pgsql/pgsql.PersistentVolumeClaim.yaml#L1-L14

chrismwendt: Added a persistent volume claim.

ggilmore:
Is this ready for review?

chrismwendt:
Yup!

ggilmore:
Are you going to add the 3.6 image tag?

chrismwendt:
Yeah, but I don't know the sha256 yet, so I'll update it once the image is pushed.

PR #346, opened 2019-08-13
update aws scaling docs

ggilmore:
- t2.xlarge only has 4 CPU cores. Some of our services require 6 cores (e.g. postgres, zoekt), so these nodes are insufficient. 

- Some quick calculations show that we need ~30 cores and ~26G RAM just for all of our services in `base/`

- `m5.4xlarge` is close to the machine type what we're using for sourcegraph.com (`n1-standard-16` - `16 cores and 60 gigs of ram`)

- k8s has a roughly equal packing (~30% requested CPU) when we use 4 `m5.4xlarge` nodes for the base deployment (no tuning)

### DIFF
--- a/docs/k8s.eks.md
+++ b/docs/k8s.eks.md
@@ -53,12 +53,9 @@ Follow the [EKS Getting Started guide](https://docs.aws.amazon.com/eks/latest/us
 
 | Users        | Instance type | Min nodes | Max nodes | Cost est.  | Attached Storage | Root Storage |
 | ------------ | ------------- | --------- | --------- | ---------- | ---------------- | ------------ |
-| 10-25        | t2.xlarge     | 4         | 6         | $18-27/day | 500 GB           | 50 GB        |
-| 25-500       | t2.xlarge     | 5         | 8         | $22-36/day | 500 GB           | 50 GB        |
-| 500-2,000    | t2.xlarge     | 6         | 8         | $27-36/day | 500 GB           | 50 GB        |
-| 2,000-4,000  | t2.xlarge     | 8         | 10        | $36-45/day | 900 GB           | 50 GB        |
-| 4,000-10,000 | t2.xlarge     | 12        | 14        | $53-62/day | 900 GB           | 50 GB        |
-| 10,000+      | t2.2xlarge    | 8         | 10        | $71-89/day | 900 GB           | 50 GB        |
+| 10-25        | m5.4xlarge    | 4         | 7         | $80-137/day | 500 GB           | 100 GB        |
+| 25-500       | m5.4xlarge   | 6         | 10         | $118-195/day | 500 GB           | 100 GB        |
+
 
 </div>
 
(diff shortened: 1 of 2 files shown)


docs/k8s.eks.md
```diff
@@ -53,12 +53,9 @@ Follow the [EKS Getting Started guide](https://docs.aws.amazon.com/eks/latest/us
...
 | Users        | Instance type | Min nodes | Max nodes | Cost est.  | Attached Storage | Root Storage |
 | ------------ | ------------- | --------- | --------- | ---------- | ---------------- | ------------ |
-| 10-25        | t2.xlarge     | 4         | 6         | $18-27/day | 500 GB           | 50 GB        |
-| 25-500       | t2.xlarge     | 5         | 8         | $22-36/day | 500 GB           | 50 GB        |
-| 500-2,000    | t2.xlarge     | 6         | 8         | $27-36/day | 500 GB           | 50 GB        |
-| 2,000-4,000  | t2.xlarge     | 8         | 10        | $36-45/day | 900 GB           | 50 GB        |
-| 4,000-10,000 | t2.xlarge     | 12        | 14        | $53-62/day | 900 GB           | 50 GB        |
-| 10,000+      | t2.2xlarge    | 8         | 10        | $71-89/day | 900 GB           | 50 GB        |
+| 10-500        | m5.4xlarge    | 3         | 6         | $59-118/day | 500 GB           | 100 GB        |
+| 500-2000       | m5.4xlarge   | 6         | 10         | $118-195/day | 500 GB           | 100 GB        |
```
slimsag: This doesn't look right to me.

m5.4xlarge is 16 CPU, 64GB RAM. What you've written here would imply that 96 vCPU 384 GB memory can only serve 500-2000 Sourcegraph users, but in https://github.com/sourcegraph/deploy-sourcegraph/blob/master/docs/scale.md#cluster-resource-allocation-guidelines we indicate 10,000+ for a lesser ammount. Uber also has more users than that with less resources IIRC.

slimsag:
Yesterday I spent some time tracking down the exact scaling specs of one of our larger customers, and am compiling a spreadsheet with this info also for other customers so we can begin offering better guidelines. https://docs.google.com/spreadsheets/d/1in1sfEkgXGVB2_HInX93bxNOFJPA_r3ugfD5lEKCk_U/edit#gid=0

Concluding thoughts on this exact PR here:

- 👍to using `m5.4xlarge` by default, I agree that is the right choice.
- Your 500-2000 recommendation here is min. six m5.4xlarge, which is 96 vCPU and 384 RAM. According to my chart above, 99 vCPU and 232 GB would be required for ~1300 users (assuming completely optimal packing). I agree your vCPU count is accurate, but the memory is very likely over-allocated (not the end of the world).

Given this is a clear improvement over what is written down today, 👍 LGTM

I plan to follow up later tonight or tomorrow with better guidelines, this is good enough for now though

ggilmore:
cc @slimsag, I went ahead and merged this

PR #355, opened 2019-08-16
Update zoekt docker images to semver tags

nicksnyder:
See https://github.com/sourcegraph/infrastructure/pull/1602


nicksnyder:
Merging without review because this doesn't change the commit that the image tags are based on.

PR #367, opened 2019-08-29
add basic integration test

ggilmore:
Screen Shot 2019-08-30 at 4 37 46 PM

This PR adds an integration test that does the following on every non-PR commit:

- creates a GKE cluster with the min spec that we have in our docs
- deploys the Sourcegraph Kubernetes yaml to it (implicitly checking to see if it's valid and if the cluster can support it)
- tests to see if the instance is reachable (response code is expected to be 401 since you need to create an admin account)

The full test takes 15-18 minutes to complete (and it's not possible to speed up since most of the time is spent spinning up / tearing down the cluster)


ggilmore:
cc @sourcegraph/distribution - this is now ready for review.

ggilmore:
One unaddressed problem - this test leaves behind some leftover disks (that were backing the PVCs). We can probably use GCP's Go SDK along with [quickTest.AddCleanup()](https://godoc.org/github.com/frankban/quicktest#C.AddCleanup) to remove them in the future.

ggilmore:
I am going to go ahead and merge this. @beyang @slimsag feel free to review after the fact

PR #375, opened 2019-09-12
frontend grafana dashboards in k8s

uwedeportivo:
this is connected to 

https://github.com/sourcegraph/sourcegraph/pull/5543

this is how this looks like in the k8s deploy context

### DIFF
--- a/base/frontend/sourcegraph-frontend.Deployment.yaml
+++ b/base/frontend/sourcegraph-frontend.Deployment.yaml
@@ -50,7 +50,7 @@ spec:
           value: /mnt/cache/$(POD_NAME)
         - name: GRAFANA_SERVER_URL
           value: http://grafana:30070
-        image: index.docker.io/sourcegraph/frontend:3.8-internal-uwe
+        image: index.docker.io/sourcegraph/frontend:3.8.0-rc.1@sha256:e579effe2e158b651704cb34cd2a8e9e52dac9dcc40843d17bd854a72a159b63
         terminationMessagePolicy: FallbackToLogsOnError
         livenessProbe:
           httpGet:
--- a/base/frontend/sourcegraph-frontend.Deployment.yaml
+++ b/base/frontend/sourcegraph-frontend.Deployment.yaml
@@ -48,7 +48,7 @@ spec:
         # archives of repositories at a commit.
         - name: CACHE_DIR
           value: /mnt/cache/$(POD_NAME)
-        - name: SRC_GRAFANA_SERVER_URL
+        - name: GRAFANA_SERVER_URL
           value: http://grafana:30070
         image: index.docker.io/sourcegraph/frontend:3.8-internal-uwe
         terminationMessagePolicy: FallbackToLogsOnError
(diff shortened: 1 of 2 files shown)


base/frontend/sourcegraph-frontend.Deployment.yaml
```diff
@@ -48,7 +48,9 @@ spec:
         # archives of repositories at a commit.
         - name: CACHE_DIR
           value: /mnt/cache/$(POD_NAME)
-        image: index.docker.io/sourcegraph/frontend:3.7.1@sha256:424bb8616fe7d042c77e91ab862fe9005c715badf6c656d13245f79ff0d03a1b
+        - name: GRAFANA_SERVER_URL
+          value: http://grafana:30070
+        image: index.docker.io/sourcegraph/frontend:3.8-internal-uwe
```
slimsag: Once you merge https://github.com/sourcegraph/sourcegraph/pull/5543 we can do an release candidate for this so it's not an `-internal-uwe` image. I can help / show you how to do that, then we can merge this PR after it is bumped.

base/grafana/grafana.Deployment.yaml
```diff
@@ -24,8 +24,21 @@ spec:
...
+        - name: GF_AUTH_ANONYMOUS_ORG_NAME
+          value: "Main Org."
+        - name: GF_AUTH_ANONYMOUS_ORG_ROLE
+          value: "Editor"
+        - name: GF_USERS_ALLOW_SIGN_UP
+          value: "false"
+        - name: GF_USERS_AUTO_ASSIGN_ORG
+          value: "true"
+        - name: GF_USERS_AUTO_ASSIGN_ORG_ROLE
+          value: "Editor"
```
slimsag: It would be ideal to move these into the container itself now that we're using our own. i.e. in the Dockerfile using:

```
ENV GF_AUTH_ANONYMOUS_ENABLED true
...
```

The reason we prefer this is to make our YAML simpler, and it becomes easier to keep this repo and other repos deploying this container in sync.

base/grafana/grafana.Deployment.yaml
```diff
@@ -24,8 +24,21 @@ spec:
         app: grafana
     spec:
       containers:
-      - image: index.docker.io/sourcegraph/grafana:6.3.3@sha256:2f68b9b1542e7d75459d983b606d2fdd1c11a75610464e3d7a6ced4f3ac474bf
+      - image: index.docker.io/sourcegraph/grafana:6.3.3-1-internal-uwe
```
slimsag: user a real published image tag here before merging

base/grafana/grafana.Deployment.yaml
uwedeportivo: that's a good idea. if it's ok with you i'll do this in a subsequent PR because i have a couple of changes i like to make to the grafana image and i'll include this then. i'll clean up the envs afterwards

base/frontend/sourcegraph-frontend.Deployment.yaml
uwedeportivo: done

base/grafana/grafana.Deployment.yaml
uwedeportivo: done

uwedeportivo:
tested on sg-uwe-4 cluster

PR #381, opened 2019-09-18
docs/migrate: remove outdated 3.7 migration information and link to docs

slimsag:
This information is now outdated and, if updated, would just restate what is already stated in the linked migration guide. So, instead, let's just link to that migration guide.


slimsag:
Looks like the integration tests on master are failing, but not due to this docs change. Merging as-is.

PR #386, opened 2019-09-30
Note docs about secondary lsif database.

efritz:
This updates the postgres config for lsif server after https://github.com/sourcegraph/sourcegraph/pull/5740.


efritz:
Note: This added and then removed PG variables as the LSIF server/worker now reads config from the frontend.

PR #388, opened 2019-10-02
declare zoekt web-server as prom target

uwedeportivo:
related to

https://github.com/sourcegraph/sourcegraph/issues/5243

depends on 

https://github.com/sourcegraph/zoekt/pull/31

landing first


uwedeportivo:
Screen Shot 2019-10-04 at 12 51 29 PM

tested that prometheus sees it now

PR #389, opened 2019-10-03
Update syntect-server to HA implementation (runs 4 workers internally)

slimsag:
- This PR updates syntect-server to the new HA implementation which internally runs 4 workers.
- Each worker can at peak use ~1.1 GiB, so a 6G memory limit is now suggested.
- The number of workers can be tuned using the env var `WORKERS`. Setting it to `1` would retain the old resource requirements but still be more stable. In practice, tuning it should not be needed.
- A Prometheus metric is also exposed by the container on `:6060/metrics` called `syntect_server_hss_worker_restarts` to measure worker restarts (any restarts indicate a bug in syntect_server).
- For more details, see the first commit message in https://github.com/sourcegraph/syntect_server/pull/25

Closes sourcegraph/sourcegraph#5406

### DIFF
--- a/base/syntect-server/syntect-server.Deployment.yaml
+++ b/base/syntect-server/syntect-server.Deployment.yaml
@@ -25,7 +25,7 @@ spec:
     spec:
       containers:
       - env:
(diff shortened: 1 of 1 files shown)


keegancsmith:
6GB for syntax highlighting, doesn't that seem a bit extreme? Is there not something else we can do, like per request limit on memory usage and just fail highlighting if it goes over?

PR #390, opened 2019-10-03
indexed-search: Migrate to a StatefulSet

keegancsmith:
Includes documentation on how to retain the existing volume.

Note: We do not yet give information on how to scale indexed-search. This will
come once horizontal sharding support is in master. Until then a statefulset
of size 1 will act in the same way as the previous deployment with size 1.

Test Plan: Tested on minikube via upgrading an older cluster.

Part of https://github.com/sourcegraph/sourcegraph/issues/5725


keegancsmith:
ping @sourcegraph/distribution

beyang:
Can you also add a note about it to the CHANGELOG in the main repo?

keegancsmith:
> LGTM, thanks for writing the migration documentation. Is there an accompanying change that we need for sourcegraph/frontend or the searcher?

No not yet. I will ship that once we are ready to enable horizontal scaling by default. But for now the existing service stays the same so there is no difference from frontend's perspective.

> Can you also add a note about it to the CHANGELOG in the main repo?

Will do.

keegancsmith:
I will run through this migration guide on sourcegraph.com and k8s.sgdev today.

PR #391, opened 2019-10-03
lsif-server: add init container to correct file permissions

slimsag:
Depends on https://github.com/sourcegraph/sourcegraph/pull/5852 but reviewable independently.

### DIFF
--- a/base/lsif-server/lsif-server.Deployment.yaml
+++ b/base/lsif-server/lsif-server.Deployment.yaml
@@ -63,6 +63,17 @@ spec:
           name: lsif-storage
       securityContext:
         runAsUser: 0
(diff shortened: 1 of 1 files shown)


base/lsif-server/lsif-server.Deployment.yaml
```diff
@@ -63,6 +63,17 @@ spec:
           name: lsif-storage
       securityContext:
         runAsUser: 0
+      # lsif-server incorrectly previously ran as `root` instead of the
+      # `sourcegraph` user. We use this init container (ran as root) to fix
```
creachadair: Completely trivial grammatical nit.
```suggestion
      # `sourcegraph` user. We use this init container (run as root) to fix
```

PR #392, opened 2019-10-03
prometheus image update (adds user)

uwedeportivo:
see https://github.com/sourcegraph/sourcegraph/pull/5855


uwedeportivo:
tested on new k8s cluster and as an upgrade from cluster running 3.8

PR #395, opened 2019-10-08
upgrade grafana

uwedeportivo:
- moves default port to 3370
- upgrades underlying grafana to 6.4.1
- fixes jsonnet dashboard generation for single server
- allows grafana to work outside the reverse proxy

related to https://github.com/sourcegraph/sourcegraph/pull/5910


uwedeportivo:
> but don't you also need to update the URL in the envvar on sourcegraph-frontend?
no, it uses the service port

PR #398, opened 2019-10-14
add prom scraping annotation to lsif

uwedeportivo:
this scrapes both ports but that's ok because both expose /metrics (meaningful and different)

we're lucky that way (see https://github.com/prometheus/prometheus/issues/3756)


efritz:
🙇

uwedeportivo:
tested on test cluster

uwedeportivo:
reverted back to last semver for frontend and lsif images so renovate can push the rc being cut tonight

PR #405, opened 2019-10-16
Remove all resource requests and require resource limits.

nicksnyder:
**What**

This PR ensures that resource requests match resource limits for all deployments. When a deployment is configured with a limit and not a request, [the request defaults to the limit](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/#what-if-you-specify-a-container-s-request-but-not-its-limit).

Many of our existing deployments already had requests===limits, so I just removed the requests config.

For deployments where requests!==limits, I set the limit to the requests. I think this is a safer change to make because it doesn't increase the amount of resources required to deploy the yaml to a fresh/minimal cluster. I think it is better to start small and have customers scale up as necessary than vice versa.

Separately, we can tune our actual allocations on dot-com based on the new dashboard that I created:
https://app.google.stackdriver.com/dashboards/16297848076041231910?project=sourcegraph-dev&timeDomain=6h

I think it is easier to tune a single toggle (e.g. requests or limits) rather than both (unless we agree to a policy like limits are always 2x requests).

**Why**

The primary benefit, and motivation, of ensuring requests===limits is that it reduces variability in performance. Each deployment is guaranteed to have a certain amount of resources and its performance will not vary based on external circumstances like other pods scheduled on the same node.

A potential future secondary benefit, is that this would make it possible for us (and customers) to more easily use a [vertical pod autoscaler](https://cloud.google.com/kubernetes-engine/docs/concepts/verticalpodautoscaler) (or equivalent) in the future, which tunes the requests (not limits) for deployments. By removing explicit requests from the yaml, applying new yaml shouldn't overwrite any modifications that a hypothetical pod autoscaler might make.


.buildkite/verify-deployment-limits.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -e
+cd "$(dirname "${BASH_SOURCE[0]}")/.."
+
+if [[ $(grep -r --include=\*.Deployment.yaml -L limits: .) ]]; then
```
keegancsmith: we should also do StatefulSets and DaemonSets

nicksnyder: Oops, good point. Will add.

slimsag:
@keegancsmith you quoted the PR description but said nothing

nicksnyder:
@keegancsmith's comment got swallowed somehow so we had a short Slack convo and posting the thoughts here.

He is worried that limits cause unreliability if they are set too low, and we request too many resources if they are set too high. Should we standardize on requests instead (and not have limits)?

My response:
If we only use requests and if we make a change that causes a service to require 2x the memory, a dev might not know to update the config. Things might be fine for a bit if the pod is on a node with extra memory, but then we are one reschedule away from the pod crashing (which could happen at any time). I am wondering if it is operationally easier to “fail fast” when resource requirements change for a deployment.

I am interested to hear what others think.

nicksnyder:
Keegan was also worried that the defaults were too low, so perhaps it would be better for me to delete the requests instead of deleting the limits and promoting the requests to limits.

nicksnyder:
Thanks for the comments. I am going to take a step back and work on an RFC so we can discuss there.

ggilmore:
@nicksnyder I talked with @slimsag and I agree with every one of his points. I'll wait for your RFC for further discussion, but:

1. I can't approve this PR as written
1. Removing resource `limits` is asking for a lot of problems
1. I think PR isn't a viable solution for the issues listed in the parent comment 
1. Being able to properly guide customers about how to tune the deployment's resource usage will require investment in testing infra and / or service re-architecture (but this is still definitely what we should be doing)

nicksnyder:
Closing this in favor of discussing [RFC 51](https://docs.google.com/document/d/1WI830xTWgHRYQs_H7g8dDLOCBbscZjY35JxxDZT7mxM/edit#).

PR #407, opened 2019-10-16
Automerge updating docker image tags

nicksnyder:
Is there any reason for changes like this to not get merged automatically? https://github.com/sourcegraph/deploy-sourcegraph/pull/406

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -14,10 +14,7 @@
       "versionScheme": "semver",
       "ignoreUnstable": false,
       "semanticCommits": false,
-      "reviewers": [
-        "ggilmore"
-      ],
-      "extends": ["default:automergeDigest"]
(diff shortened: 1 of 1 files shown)


ggilmore:
@nicksnyder There have been situations in the past where someone has overwritten one of our git tags (e.g. tagging two different commits as `3.x.x.`). If these updates are automerged, then it's very hard for us to detect this.

nicksnyder:
I don't think I understand your concern. Can you elaborate what the situation is and how merging changes like this make?

> There have been situations in the past where someone has overwritten one of our git tags

Git tag in this repo or in sourcegraph/sourcegraph?

> If these updates are automerged, then it's very hard for us to detect this.

What happens differently if this is merged or not?

PR #408, opened 2019-10-16
add script to delete unattached GCP "Sourcegraph Auxiliary" disks 

ggilmore:
- We have a limited disk quota on the Sourcegraph Auxiliary project, so unused disks should be removed promptly 

- Nobody should be creating long-lived services in the Sourcegraph Auxiliary project, so it's safe to delete these disks. (You should create a snapshot if you want to retain data after you delete an instance)

- This script should ultimately run as a cronjob on our CI cluster, but that requires some additional configuration. Adding it as a simple script in this repo is faster, and is still useful since this repository's integration tests create many of these disks.


.buildkite/cleanup-unused-disks.sh
```diff
@@ -0,0 +1,20 @@
...
+
+PROJECT=sourcegraph-server
+
+gcloud_command() {
+    gcloud --quiet --project="$PROJECT" "$@"
+}
+
+echo "--- Deleting unattached GCP disks from the $PROJECT project"
+
+for disk in $(gcloud_command compute disks list --filter="-users:*" --format="value(selfLink)")
```
beyang: Does this *only* match unattached disks? When I ran this against the `sourcegraph-dev` project, it included disks like `pgsql-prod-dot-com` and `zoekt-data`, which seem very much attached and critical.

ggilmore: Screen Shot 2019-10-16 at 4 17 56 PM
@beyang Those disks that you mentioned are indeed unattached.

ggilmore: @beyang Additionally, ["Disks can be deleted only if they are not being used by any virtual machine instances"](https://cloud.google.com/sdk/gcloud/reference/compute/disks/delete).

beyang: worth calling out in a comment, I think. Otherwise, LGTM.

PR #430, opened 2019-10-30
prometheus: config file from ConfigMap

uwedeportivo:
This moves the k8s prometheus config into the ConfigMap providing full external control without the need to rev the prometheus disk image.


uwedeportivo:
tested on gcp test cluster

PR #431, opened 2019-10-30
prometheus: constrain service scraping to sourcegraph.prometheus

uwedeportivo:
This changes the scraping annotation for services to sourcegraph.prometheus/scrape.

This makes it less likely that customers with additional prometheus instances running in their cluster will have the same label and it prevents our builtin prometheus from scraping services not related to Sourcegraph. The previous annotation is probably what everybody uses because it's copy/pasted from the prometheus kubernetes docs.

Note: this doesn't address the other scraping roles: nodes, endpoints and pods.

We still want those for our dot-com deployment but I'm not sure how useful they are in a customer deployment context. We can restrict those too (in the same way) or delete them.

### DIFF
--- a/base/frontend/sourcegraph-frontend.Service.yaml
+++ b/base/frontend/sourcegraph-frontend.Service.yaml
@@ -3,7 +3,7 @@ kind: Service
 metadata:
   annotations:
     prometheus.io/port: "6060"
-    prometheus.io/scrape: "true"
+    sourcegraph.prometheus/scrape: "true"
   labels:
     app: sourcegraph-frontend
     deploy: sourcegraph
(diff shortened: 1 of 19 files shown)


slimsag:
> pods/endpoints -> I'm not sure what metrics we get from these, I agree it'd be nice to keep them on dot-com but probably not important for customers
>
> nodes -> this is the only way we get node-level CPU/memory/disk metrics, right? we would want to keep those in customer environments for sure

uwedeportivo:
for nodes the simplest solution would be to run sourcegraph in a namespace and limit scraping of nodes in that namespace with 

(for example namespace foo):

```
      - source_labels: [__meta_kubernetes_namespace]
        action: keep
        regex: foo
```

PR #443, opened 2019-11-13
alert: gitserver low disk

uwedeportivo:
Work for issue:

https://github.com/sourcegraph/sourcegraph/issues/5994

This sets up alerts according to what the gitserver processes see through syscall stats exported as prometheus metrics.

### DIFF
--- a/base/prometheus/prometheus.ConfigMap.yaml
+++ b/base/prometheus/prometheus.ConfigMap.yaml
@@ -290,6 +290,18 @@ data:
               help: Alerts when a node has less than 5% available disk space.
               summary: Critical! {{`{{`}}$labels.exported_name{{`}}`}} has less than 5% available
                 disk space
+          - alert: GitserverDiskSpaceLow
+            expr: src_gitserver_disk_space_available / src_gitserver_disk_space_total < 0.1
+            annotations:
+              help: Alerts when gitserverdisk space is low.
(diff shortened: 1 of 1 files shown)


base/prometheus/prometheus.ConfigMap.yaml
```diff
@@ -290,6 +290,18 @@ data:
...
+            expr: src_gitserver_disk_space_available / src_gitserver_disk_space_total < 0.1
+            annotations:
+              help: Alerts when gitserverdisk space is low.
+              summary: gitserver {{`{{`}}$labels.instance{{`}}`}} disk space is less than 10% of available disk space
+          - alert: GitserverDiskSpaceLowCritical
+            expr: src_gitserver_disk_space_available / src_gitserver_disk_space_total < 0.05
+            labels:
+              severity: page
+            annotations:
+              help: Alerts when gitserverdisk space is critically low.
```
ggilmore: Note that the cleanup threshold is actually set by an environment variable: https://github.com/sourcegraph/sourcegraph/blob/7da2f8d8a3851083090d7c6bf4d3d3dc35745697/cmd/gitserver/main.go#L30

So these hard-coded alerts may not trigger when you expect them to.

uwedeportivo: i made myself an issue https://github.com/sourcegraph/deploy-sourcegraph/issues/456

uwedeportivo:
i'm on pager duty this week and gitlab integration is happening, so i'll land this one

PR #447, opened 2019-11-15
Specify postgresql.conf in ConfigMap

beyang:
This makes it so the `postgresql.conf` file is specified in a ConfigMap, so that the site admin can customize the PostgreSQL config.

### DIFF
--- a/base/pgsql/pgsql.Deployment.yaml
+++ b/base/pgsql/pgsql.Deployment.yaml
@@ -26,7 +26,7 @@ spec:
     spec:
       containers:
       - env:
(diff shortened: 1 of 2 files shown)


base/pgsql/pgsql.Deployment.yaml
```diff
@@ -51,6 +51,8 @@ spec:
         volumeMounts:
         - mountPath: /data
           name: disk
+        - mountPath: /conf
```
tsenart: Postgres expects the config file in $PGDATA. I don't see immediately how we're telling it to look for the config file in another place. What am I missing?

uwedeportivo: this: https://github.com/sourcegraph/infrastructure/pull/1728

tsenart: Thanks @uwedeportivo!

PR #465, opened 2019-12-03
gitserver: app label for service

uwedeportivo:
prometheus needs the app label.

i added it to the selector too. in the next iteration we can remove the type labels from the statefulset and service if we want to make it more uniform.


uwedeportivo:
i did, i tested on dot-com actually (blush)

uwedeportivo:
i need a test cluster anyways, i'll test it one more time

uwedeportivo:
tested on test cluster too

PR #513, opened 2020-01-30
Namespaced sourcegraph

uwedeportivo:
version of our k8s configs inside a namespace and with no cluster role creation requirements


base/frontend/sourcegraph-frontend.RoleBinding.yaml
```diff
@@ -5,10 +5,12 @@ metadata:
     category: rbac
     deploy: sourcegraph
   name: sourcegraph-frontend
+  namespace: ns-sourcegraph
 roleRef:
   apiGroup: ""
-  kind: Role
-  name: sourcegraph-frontend
+  kind: ClusterRole
+  name: view
```
keegancsmith: is this a predefined k8s clusterrole?

base/frontend/sourcegraph-frontend.RoleBinding.yaml
```diff
@@ -5,10 +5,12 @@ metadata:
...
 subjects:
 - kind: ServiceAccount
   name: sourcegraph-frontend
+  namespace: ns-sourcegraph
```
keegancsmith: is this the magic? :)

base/frontend/sourcegraph-frontend.RoleBinding.yaml
uwedeportivo: yes

base/frontend/sourcegraph-frontend.RoleBinding.yaml
uwedeportivo: i guess :-)

uwedeportivo:
just to clarify @keegancsmith: this is not ready for merging. we need to figure out some things like docs, migration plans etc

this was more a proof of concept to see if we can live with reduced cluster privileges

keegancsmith:
> just to clarify @keegancsmith: this is not ready for merging. we need to figure out some things like docs, migration plans etc
> 
> this was more a proof of concept to see if we can live with reduced cluster privileges

cool. Yeah I think this is a good move in general. We have the optional daemonset pod-tmp-gc which does need node permissions. But that is optional. Then of course we have some node based monitoring in prometheus (I see you tackle that here). But nothing fundamentally needs anything more than getting the pods behind a service. 
If that becomes an issue, we can look into using non-k8s api service discovery. May require changing the searcher (and other services) to a headless service. Or even switching those to statefulsets and stop relying on host volumes. Probably a good thing in general to use more generic service discovery mechanisms anyway. Ok thats enough verbal diarrhoea for me this early in the morning.

slimsag:
Merged `master` into this branch, effectively bringing it up to v3.13.1.

Also created instructions for anyone who wishes to use this branch today: https://docs.google.com/document/d/1vEJDgosKiBwfK8OWmUtRccQrnpcR3Ziq-PnQLKfN5Bk/edit?usp=sharing

uwedeportivo:
not needed anymore, superseded by https://github.com/sourcegraph/deploy-sourcegraph/pull/539

PR #521, opened 2020-02-10
Use cue as config language

uwedeportivo:
**Proof of Concept**

Use [cuelang](https://cuelang.org/) to define the k8s configs.

> NOTE: I left the yaml files in there for easier comparison. They are not needed anymore so if we do this for real we can delete them.

> NOTE: this needs the latest cue, so head of master from https://github.com/cuelang/cue.

I started with the yaml files from base, imported them into cue and then lifted common stuff into base.cue and started defining constraints. 

base.cue contains the boilerplate and the types/constraints and the individual cue files in the services are specializations of the structs in base.cue.

Advantages:

* we can define constraints that all our Sourcegraph installations need to respect
* we can import and define common k8s constraints (for example that port definitions match)
* we can reduce boilerplate
* we can have use cue to generate useful specializations:
     * change of namespace (or no namespace)
     * run as root or nonRoot
     * different volume providers
     * run on different k8s runtimes (ie minikube)

Disadvantages

* yet another config language
* fairly new (although istio is using it now and it has traction inside Google)

The cue docs are really good, the project is active and the codebase is Go and relatively straightforward (so we could contribute).

Here are the commands I used:

```
cd base
cue import ./... -p base -l 'strings.ToCamel(kind)' -l metadata.name -f -R
```

Then I extracted structs and fields into base.cue and checked my progress with

```
cue eval -c ./... > snapshot
```

to make sure nothing got lost in the process. After that I trimmed the specialized cue files

```
cue trim ./...
```

There is still work to be done in this lifting loop. Volume claims and volume claim templates can be constrained, port definitions can be tightened up. But you get the idea. If we get down this road I can trim more.

Once that is done you can generate yaml files that look just like the existing ones or you can feed the structs directly into k8s (cue can drive like kubectl).


keegancsmith:
<3 I experimented with converting this repo to cue a while ago and was very impressed with the tooling, the model and docs.

I suppose to move forward with this we would want most users to not need to interact with cue, but instead just work on the committed generated files? And then when users need to do more modifications/etc, they then switch to using cue? I am guessing those on the distribution team would have a good feeling for what new customers are willing to do :)

base/base.cue
```diff
@@ -0,0 +1,159 @@
...
+	}
+	spec: {
+		// 1 is the default, but we allow any number
+		replicas: *1 | int
+		template: {
+			metadata: labels: {
+				app: *ID | string
+			}
+			// we always have one namesake container
+			spec: containers: [{name: *ID | string}, ...]
```
nicksnyder: Is this always true?

uwedeportivo: i relaxed that constraint, i will delete the comment.

slimsag:
I just looped back to this now, there is a Slack thread with some additional conversation around Cue vs. Dhall etc. https://sourcegraph.slack.com/archives/CJX299FGE/p1582586025017300

PR #523, opened 2020-02-11
DO NOT MERGE - fix pgsql /conf user permissions

uwedeportivo:
fix to unblock james

not intended to be merged. we will have a real fix for this situation soon but it involves building a new pgsql docker image.


uwedeportivo:
this was a hack that's not needed anymore

PR #528, opened 2020-02-19
nginx-ingress: upgrade to 0.29.0

ggilmore:
Previously, our nginx pods were failing with the following error: 

```
Failed to list *v1beta1.Ingress: ingresses.networking.k8s.io is forbidden: User "system:serviceaccount:ingress-nginx:nginx-ingress-serviceaccount" cannot list resource "ingresses" in API group "networking.k8s.io" at the cluster scope
```

### DIFF
--- a/configure/ingress-nginx/mandatory.yaml
+++ b/configure/ingress-nginx/mandatory.yaml
@@ -208,7 +208,7 @@ spec:
         app.kubernetes.io/part-of: ingress-nginx
       annotations:
         prometheus.io/port: "10254"
-        prometheus.io/scrape: "true"
+        sourcegraph.prometheus/scrape: "true"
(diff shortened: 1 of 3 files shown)


configure/ingress-nginx/mandatory.yaml
```diff
@@ -206,12 +208,16 @@ spec:
         app.kubernetes.io/part-of: ingress-nginx
       annotations:
         prometheus.io/port: "10254"
-        sourcegraph.prometheus/scrape: "true"
+        prometheus.io/scrape: "true"
```
uwedeportivo: our sourcegraph prometheus instance looks for "sourcegraph.prometheus/scrape" label

PR #532, opened 2020-02-24
use kustomize to set a namespace

uwedeportivo:
kustomize (https://kustomize.io/) is built into kubectl and seems a pretty standard way to do customizations like a namespace, custom config maps, overlays, etc.

This experiment proves that we can launch a Sourcegraph instance in a namespace like so:

```sh
cd base
# create the necessary storage class and then:
kubectl create namespace ns-sourcegraph
kubectl apply -k .
```

There's way more possibilities here with kustomization. I will try out more stuff in this direction. It is possible this will be a way for us to provide pemission-restricted k8s configs, configs for different runtimes (ie minikube or k3s), different scaling params (replicas etc). And it seems to be a blessed way to do it by the kubernetes team (integrated and documented).

Tested on a gcp cluster.

### DIFF
--- a/base/kustomization.yaml
+++ b/base/kustomization.yaml
@@ -1,0 +1,57 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: ns-sourcegraph
+resources:
+  - searcher/searcher.Service.yaml
+  - searcher/searcher.Deployment.yaml
+  - backend.Service.yaml
+  - frontend/sourcegraph-frontend.Role.yaml
+  - frontend/sourcegraph-frontend-internal.Service.yaml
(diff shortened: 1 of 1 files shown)


uwedeportivo:
[examples](https://kubectl.docs.kubernetes.io/pages/examples/kustomize.html) on what's possible

uwedeportivo:
@beyang globbing syntax is dangerous with kustomize. you have to be explicit to avoid unwanted side-effects. anyways, it got removed https://github.com/kubernetes-sigs/kustomize/issues/217, so it's not even possible :-)

uwedeportivo:
superseded by https://github.com/sourcegraph/deploy-sourcegraph/pull/541

PR #534, opened 2020-02-25
switch away from user 0 (root) and delete unnecessary cluster role

uwedeportivo:
and cluster role binding

this is stage one in our migration to accommodate restrictive cluster environments.

(stage two will allow specifying namespaces through kustomization.yaml and adapt prometheus configs)

tested both new cluster and migrating old cluster

the init containers do need to run as root to have the power to change data ownership, once that is done they can be removed.

i'm happy to hear ideas on how we can make this more seamless (would be nice to have something like how we do db migrations :-) )


uwedeportivo:
hold off the review for now. still having issues. debugging it

uwedeportivo:
the one that gives me trouble is pgsql. there's a whole matrix of situations on this one and too many variables in the entrypoint script

uwedeportivo:
all good now, tested it multiple times with existing and fresh

base/gitserver/gitserver.StatefulSet.yaml
```diff
@@ -20,6 +20,15 @@ spec:
         group: backend
         type: gitserver
     spec:
+      initContainers:
+      - name: transfer-file-ownership
+        image: sourcegraph/alpine:3.10@sha256:4d05cd5669726fc38823e92320659a6d1ef7879e62268adec5df658a0bacf65c
+        command: ["sh", "-c", "chown -R 100:100 /data/repos"]
```
keegancsmith: is there a way to skip running this if the perms are fine already? this could be a slow operation on large disks.

uwedeportivo:
rollbacks work. i tried: rollbacks from previous root -> nonroot -> root again and
 rollback from nonroot -> root

base/gitserver/gitserver.StatefulSet.yaml
uwedeportivo: well observed. i will try a couple of things. maybe we can do something here

uwedeportivo: i did it for gitserver which has big disks

docs/migrate.md
```diff
@@ -4,6 +4,39 @@ This document records manual migrations that are necessary to apply when upgradi
 Sourcegraph versions. All manual migrations between the version you are upgrading from and the
 version you are upgrading to should be applied (unless otherwise noted).
 
+## 3.14 (Unreleased)
+
+Version 3.14 changes the security context of the installation by switching to a non-root user for all containers.
+This allows running Sourcegraph in clusters with restrictive security policies.
+
+Existing installations that have been run as root before need to migrate their persistent volumes to work in 3.14.
```
keegancsmith: @uwe this needs to be called out in the CHANGELOG on sourcegraph/sourcegraph. Since if I am not mistaken if an admin doesn't do this, everything will implode :)

uwedeportivo: done

PR #543, opened 2020-03-06
frontend: lists services for /-/debug page

uwedeportivo:
related to https://github.com/sourcegraph/sourcegraph/pull/8805

### DIFF
--- a/base/frontend/sourcegraph-frontend.Role.yaml
+++ b/base/frontend/sourcegraph-frontend.Role.yaml
@@ -11,6 +11,7 @@ rules:
   - ""
   resources:
   - endpoints
+  - services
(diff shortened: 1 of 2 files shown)


base/frontend/sourcegraph-frontend.Role.yaml
```diff
@@ -11,6 +11,7 @@ rules:
   - ""
   resources:
   - endpoints
+  - services
```
keegancsmith: can we include a comment here about why we need services? (and retro-acetively why we need endpoints?)

uwedeportivo: done

PR #545, opened 2020-03-09
doc: clarify and separate in migration doc the two changes in 3.14

uwedeportivo:
Makes clear what the two changes are and which one is needed for migrating existing installations.

### DIFF
--- a/docs/migrate.md
+++ b/docs/migrate.md
@@ -39,7 +39,7 @@ restart the necessary containers.
 New installations do not need this `kustomization` and existing installations can operate from base again after the
 migration.
 
-### New installations with restrictive security policies
+### New installations: accommodate clusters with restrictive security policies
 
 New installations on clusters with restrictive security policies can now use a kustomization to accomodate those restrictions:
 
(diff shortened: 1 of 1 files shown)


docs/migrate.md
```diff
@@ -6,6 +6,8 @@ version you are upgrading to should be applied (unless otherwise noted).
 
 ## 3.14 (Unreleased)
 
+### Migrating from container user root to non-root user
```
christinaforney:
```suggestion
### Existing installations: Migrating the container user from root to non-root
```

docs/migrate.md
```diff
@@ -37,6 +39,8 @@ restart the necessary containers.
 New installations do not need this `kustomization` and existing installations can operate from base again after the
 migration.
 
+### New installations with restrictive security policies
```
christinaforney: Is this section actually a migration? Or should this just be part of our Kubernetes documentation for this specific case? 

```suggestion
### New installations: accommodate clusters with restrictive security policies
```

PR #549, opened 2020-03-17
file group ownership matches container user:group

uwedeportivo:
in PR https://github.com/sourcegraph/deploy-sourcegraph/pull/534 i made a mistake of just checking

```
id -u <container-user>
```

and assuming group would be the same id. this is just a convention not necessarily true, so this time i checked

```
id -g <container-user>
```
too and made the required changes.

I will test everything one more time (migration and new from scratch and rollback)

NOTE: this all worked before and was tested. the theory is that the fsGroup value makes it work. But having them match exactly is more desirable.

### DIFF
--- a/base/pgsql/pgsql.Deployment.yaml
+++ b/base/pgsql/pgsql.Deployment.yaml
@@ -70,8 +70,8 @@ spec:
             cpu: 10m
             memory: 50Mi
         securityContext:
-          runAsUser: 100
-          runAsGroup: 100
+          runAsUser: 999
+          runAsGroup: 999
       securityContext:
         fsGroup: 999
       volumes:
--- a/overlays/migrate-to-nonroot/redis/redis-store.Deployment.yaml
+++ b/overlays/migrate-to-nonroot/redis/redis-store.Deployment.yaml
@@ -8,7 +8,7 @@ spec:
       initContainers:
         - name: transfer-file-ownership
           image: sourcegraph/alpine:3.10@sha256:4d05cd5669726fc38823e92320659a6d1ef7879e62268adec5df658a0bacf65c
-          command: ["sh", "-c", "chown -R 999:999 /redis-data"]
+          command: ["sh", "-c", "chown -R 999:1000 /redis-data"]
           volumeMounts:
             - mountPath: /redis-data
               name: redis-data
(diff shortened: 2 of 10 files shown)


uwedeportivo:
- migration works
- starting from scratch works

testing rollback now

(i add repos and do a search to make sure things get exercised)

slimsag:
> (i add repos and do a search to make sure things get exercised)

Be sure to check containers logs. Unfortunately a lot of these services like e.g. Redis will just log errors but won't outright fail depending on what they do

slimsag:
https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#discussion explicitly states:

> fsGroup: Volumes that support ownership management are modified to be owned and writable by the GID specified in fsGroup. See the Ownership Management design document for more details.

So I think you should be 100% OK with this change.

base/pgsql/pgsql.Deployment.yaml
```diff
@@ -70,8 +70,8 @@ spec:
             cpu: 10m
             memory: 50Mi
         securityContext:
-          runAsUser: 100
-          runAsGroup: 100
+          runAsUser: 999
```
slimsag: Was the user here wrong before? Just double-checking this wasn't a typo

uwedeportivo: wasn't a typo. i checked the user "postgres" and it's 999.  it works with other users too according to the documentation. not sure what's better, but both work. i  did see this in the pgsql logs

```
Warning: you don't own a lock of type ExclusiveLock
```

but i saw this before too with user 100.

uwedeportivo:
i'm checking container logs. so far it's clean

uwedeportivo:
rollback works and logs are clean

overlays/migrate-to-nonroot/redis/redis-store.Deployment.yaml
```diff
@@ -8,7 +8,7 @@ spec:
       initContainers:
         - name: transfer-file-ownership
           image: sourcegraph/alpine:3.10@sha256:4d05cd5669726fc38823e92320659a6d1ef7879e62268adec5df658a0bacf65c
-          command: ["sh", "-c", "chown -R 999:999 /redis-data"]
+          command: ["sh", "-c", "chown -R 999:1000 /redis-data"]
```
slimsag: Oops @uwedeportivo we fixed this here but didn't update base/redis/redis-store.Deployment.yaml

PR #556, opened 2020-03-19
fix 3.13 to use correct branch

uwedeportivo:
fix 3.13.2 release error

### DIFF
--- a/base/frontend/sourcegraph-frontend.Deployment.yaml
+++ b/base/frontend/sourcegraph-frontend.Deployment.yaml
@@ -50,7 +50,7 @@ spec:
           value: /mnt/cache/$(POD_NAME)
         - name: GRAFANA_SERVER_URL
           value: http://grafana:30070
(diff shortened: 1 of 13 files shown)


rvantonder:
Does this fix the thing causing build failure on some tests?

> FAIL: TestLatestKubernetesVersionPushed (0.30s)
  | handler_test.go:45: Could not find Kubernetes release 3.13.2 on GitHub. Response code 404 Not  Found from https://github.com/sourcegraph/deploy-sourcegraph/releases/tag/v3.13.2, err: <nil>
  | FAIL

https://buildkite.com/sourcegraph/sourcegraph/builds/58656#26b12192-4315-4d0c-aa30-7b6c16c8c1ff

Stamping if so.

uwedeportivo:
thanks @rvantonder. it is part of the solution. we borked 3.13.2 in deploy-sourcegraph and we're fixing it

PR #559, opened 2020-03-22
Jaeger without the Jaeger Operator

beyang:
* Removes the Jaeger Operator in favor of including the Jaeger k8s files directly in deploy-sourcegraph. (The Jaeger Operator required permissions to create a Custom Resource Definition and added an additional layer of complexity to reason about that wasn't that useful.)
* Installs the Jaeger all-in-one instance as part of the default Sourcegraph installation.
  * `--memory.max-traces=20000` to limit memory consumption and memory/cpu resources of Jaeger resource objects are limited, but sufficient to handle 20000 spans (under the assumption that each span is less than 1KB)
* Defines Dockerfiles for jaeger-agent and jaeger-all-in-one. These build off the official images, but (1) include common unix commands for debuggability and (2) run as non-root users.
* Updates docs.

TODO
- [x] Determine if `bind-tools` is necessary (no)

Post-merge follow-up
- [ ] Note in the CHANGELOG the increased resource utilization overhead that Jaeger introduces (1/10th of a cpu, 100mb per jaeger-agent, 1/2 cpu, 500mb for the jaeger-all-in-one)
- [ ] Link to the new docs from CHANGELOG
- [ ] Jaeger for Docker Compose deployment
  - [ ] Address Stephen's comments pertaining to the Docker Compose environment.
- [ ] Jaeger for Single Docker container deployment


base/frontend/sourcegraph-frontend.Deployment.yaml
```diff
@@ -88,6 +88,33 @@ spec:
...
+        - containerPort: 5778
+          protocol: TCP
+        - containerPort: 6831
+          protocol: UDP
+        - containerPort: 6832
+          protocol: UDP
+        resources:
+          limits:
+            cpu: "1"
+            memory: 500M
```
keegancsmith, line 111: is this enough mem for the agent? Wouldn't want to introduce downtime due to the agent getting too many spans.

Is there a way to tell jaeger how much memory to use so it can throw away spans when it is using too much?

configure/jaeger/README.md
```diff
@@ -1,207 +1,57 @@
...
-   in the Kubernetes cluster.
+To disable Jaeger entirely, do:
 
-1. Deploy Jaeger as a Custom Resource Definition object:
+1. Update the Sourcegraph [site
+   configuration](https://docs.sourcegraph.com/admin/config/site_config) to remove the
+   `observability.tracing` field.
+1. Remove the `base/jaeger` directory: `rm -rf base/jaeger`
+1. Remove the jaeger agent containers from each `*.Deployment.yaml` and `*.StatefulSet.ayml` file.
+1. Apply these changes to the cluster.
```
keegancsmith, line 33: can we provide kustomizations for these? This seems quite labour intensive for our admins + asking for merge conflicts down the road.

dockerfiles/jaeger/agent/Dockerfile
```diff
@@ -0,0 +1,18 @@
+# This Dockerfile tracks jaegertracing/jaeger-agent, but includes more debugging tools and runs as a
+# non-root user. It requires JAEGER_VERSION to be set as an argument to build.
+ARG JAEGER_VERSION
+FROM jaegertracing/jaeger-agent:$JAEGER_VERSION as base
+
+FROM alpine:3.11.5
+RUN apk update
+RUN apk add bash curl
```
keegancsmith, line 8:
```suggestion
RUN apk --no-cache bash curl
```

we don't need to distribute the apk cache in the image, the above suggestion does what you want without doing that.

dockerfiles/jaeger/agent/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+COPY --from=base /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
```
keegancsmith, line 10: you could just install ca-certificates in the above line, would seem "safer". Also do we need `bind-tools` @slimsag? this agent does need to resolve the address of the jaeger collector. However, I don't think bind-tools is needed in k8s?

dockerfiles/jaeger/agent/Dockerfile
```diff
@@ -0,0 +1,18 @@
+# This Dockerfile tracks jaegertracing/jaeger-agent, but includes more debugging tools and runs as a
+# non-root user. It requires JAEGER_VERSION to be set as an argument to build.
+ARG JAEGER_VERSION
+FROM jaegertracing/jaeger-agent:$JAEGER_VERSION as base
+
+FROM alpine:3.11.5
```
keegancsmith, line 6: why do you create a new image? Why not just run commands in the jaeger-agent image?

beyang: The jaeger-agent image ([defined here](https://github.com/jaegertracing/jaeger/blob/master/cmd/agent/Dockerfile)) uses `scratch` as its base image. Consequently, it contains nothing except the agent binary itself. It doesn't have `apk` or even a shell. If we want to install debugging tools, it's much easier to start from a linux distro base image.

dockerfiles/jaeger/agent/Dockerfile
beyang: Is `bind-tools` necessary for DNS resolution? If so, it may be required for the "send spans to external jaeger-collector" use case (which is documented, but I haven't actually tested). I will check that scenario.

`apk add ca-certificates` is indeed [what is run in the standard jaeger-agent image](https://github.com/jaegertracing/jaeger/blob/master/cmd/agent/Dockerfile#L2). We could run it here instead of copying the value from the standard jaeger-agent image. Benefit would be we might have slightly more up-to-date certs, but at the cost of not as closely tracking the source image. My preference is to track the base image as closely as possible (and update the base image version if an update is required).

configure/jaeger/README.md
beyang: Kustomizations aren't composable, which means they aren't ideal for the use case of turning on and off different components in the Sourcegraph k8s distribution. In addition, `kubectl apply -k` has surprising behavior such as `--prune` no longer working, which is surprising for admins who currently don't use kustomize. For these reasons, we are actually rethinking our decision to rely more heavily on kustomize, so manual instructions with the possibility of merge conflicts seems like the best option for now.

(Aside: Helm might actually be close to what we want than kustomize, now that it no longer includes tiller.)

base/frontend/sourcegraph-frontend.Deployment.yaml
beyang: > is this enough mem for the agent? Wouldn't want to introduce downtime due to the agent getting too many spans.

I believe so. The [default jaeger-agent queue size is 1000 spans](https://www.jaegertracing.io/docs/1.17/performance-tuning/#adjust-server-queue-sizes), after which spans should be dropped in FIFO order. If the jaeger-agent exceeds its memory limit and is OOMKilled, it should be restarted without affecting the availability of the main container. If the jaeger-agent exceeds its memory limit and the pod is evicted, this could cause some downtime while waiting for the new pod to be scheduled, but this is within the "contract" of k8s (i.e., pods might be killed at any point in time, use multiple replicas to ensure reliability).

> Is there a way to tell jaeger how much memory to use so it can throw away spans when it is using too much?

Yes, the jaeger-all-in-one image has a `--memory.max-spans` parameter (which is set in this PR) to limit the maximum number of spans it stores before evicting older spans in FIFO order.

beyang:
> Are there any implications to requiring an extra 100m CPU and mem per replica in our clusters?

Yes, if a cluster is at 100% resource utilization, this might require the addition of another node to the cluster. I will call this out in the docs.

dockerfiles/jaeger/agent/Dockerfile
keegancsmith: > Is bind-tools necessary for DNS resolution?

I think at some customers using the deploy-sourcegraph-docker method needed the bind-tools package. I don't think it is needed for k8s clusters, but just wanted to confirm that.

> If so, it may be required for the "send spans to external jaeger-collector" use case (which is documented, but I haven't actually tested). I will check that scenario.

Isn't it also needed to send to the collector in the cluster? IE we do a dns lookup, don't use a hardcoded IP.

> apk add ca-certificates is indeed what is run in the standard jaeger-agent image . We could run it here instead of copying the value from the standard jaeger-agent image. Benefit would be we might have slightly more up-to-date certs, but at the cost of not as closely tracking the source image. My preference is to track the base image as closely as possible (and update the base image version if an update is required).

Yeah sounds good to just keep this line as it is then.

beyang: > Isn't it also needed to send to the collector in the cluster? IE we do a dns lookup, don't use a hardcoded IP.

I don't think so? At least, the jaeger-agent has no problem talking to the collector in the test cluster I spun up.

dockerfiles/jaeger/agent/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+RUN apk --no-cache bash curl
```
slimsag, line 7: Are these Dockerfiles ONLY going to be used by deploy-sourcegraph, or are they also going to be used by deploy-sourcegraph-docker?

If the latter, these should all be moved into https://github.com/sourcegraph/sourcegraph/tree/master/docker-images and follow the conventions there.

This is also extremely important because I want us to get all Docker images versioned alongside Sourcegraph (`3.14.1` instead of `<arbitrary>` tags).

dockerfiles/jaeger/agent/Dockerfile
slimsag: `bind-tools` is needed for DNS resolution in Docker environments. See https://sourcegraph.com/github.com/sourcegraph/sourcegraph@730f31fb13b8380ed02fd01fe4a8583e43b029d4/-/blob/docker-images/alpine/Dockerfile#L17-23

dockerfiles/jaeger/all-in-one/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+# Collector HTTP
+EXPOSE 14268
+
+# Collector gRPC
+EXPOSE 14250
+
+# Web HTTP
+EXPOSE 16686
+
+VOLUME ["/tmp"]
```
slimsag, line 40: Will this Docker image ever be used in non-Kubernetes environments? If yes, you MUST create `/tmp` within the image with the desired permissions or else the container will not be able to write to `/tmp` in certain environments. See https://sourcegraph.com/github.com/sourcegraph/sourcegraph@730f31fb13b8380ed02fd01fe4a8583e43b029d4/-/blob/cmd/precise-code-intel/api-server/Dockerfile#L30-37

beyang:
I will address @slimsag's comments in a follow up to this PR (as I tackle Jaeger in Docker Compose).

PR #568, opened 2020-03-27
docs: migrate explains harmless error message with kubectl-apply-all.sh

uwedeportivo:
apply -f to a whole directory complains when finding a kustomization.yaml file.

i screwed this up, sorry. it's a harmless message which unfortunately is displayed as an error.

we could make them switch to apply -k by changing kubectl-apply-all.sh to use that instead. the downside is that we then require kubectl client version >= 1.14, which is probably acceptable since it's more than a year old.


uwedeportivo:
this was solved in a different way (we feed apply -f explicit list of files avoiding kustomization.yaml)

PR #569, opened 2020-03-27
Add explicit lsif/precise-code-intel service envvars

efritz:


base/frontend/sourcegraph-frontend.Deployment.yaml
```diff
@@ -50,6 +50,12 @@ spec:
           value: /mnt/cache/$(POD_NAME)
         - name: GRAFANA_SERVER_URL
           value: http://grafana:30070
+          # old value during transition
+        - name: LSIF_API_SERVER_URL
+          value: k8s+http://lsif-server:3186
+          # new value after transition
```
slimsag: This YAML isn't going to be automatically deployed until you merge it into the upstream deploy-sourcegraph-dot-com; so you do not need to have this old value for any transitional phase. Also realize that customers would also not have this.

slimsag:
Uhg, I am an idiot and am wrong here. Never-mind. This LGTM as-is.

PR #570, opened 2020-03-27
integration tests: test deploying as restricted admin

uwedeportivo:
captured the steps as a shell script.

still to be done:

- not sure how to capture it in pulumi (the fake user impersonation part for example or kustomization overlays)

this will close out RFC 103


uwedeportivo:
https://github.com/kubernetes-sigs/kustomize/blob/master/docs/eschewedFeatures.md

uwedeportivo:
i like pulumi but i have to say the non-pulumi version was 4 min faster. maybe this run just got lucky and had a better buildkite node

uwedeportivo:
ready for review

kubectl-apply-all.sh
```diff
@@ -6,5 +6,6 @@
 #   * When the cluster is first created
 #   * Whenever the configuration for any resource has been updated
 
-# Apply the base Soucegraph deployment 
-kubectl apply --prune -l deploy=sourcegraph -f base --recursive
```
keegancsmith: the removal of prune means we no longer will remove no longer used services?

tests/integration/restricted/test.sh
```diff
@@ -0,0 +1,56 @@
...
+
+RANDOM_CLUSTER_NAME_SUFFIX=`cat /dev/urandom | tr -cd 'a-f0-9' | head -c 7`
+
+CLUSTER_NAME="ds-test-restricted-${RANDOM_CLUSTER_NAME_SUFFIX}"
+
+cd $(dirname "${BASH_SOURCE[0]}")
+
+# set up the cluster, set up the fake user and restricted policy and then deploy the non-privileged overlay as that user
+
+gcloud container clusters create ${CLUSTER_NAME} --zone ${TEST_GCP_ZONE} --num-nodes 3 --machine-type n1-standard-16 --disk-type pd-ssd --project ${TEST_GCP_PROJECT} --labels=cost-category=build
```
keegancsmith: I think some more labels here would be cool, extracted from the buildkite env. Mostly interested in the user attached to the build. Could pick up things like a bot causing lots of builds / etc. Another interesting one would be if its for master vs a PR.

Additionally a label sayings this is a deploy-sourcegraph integration test would be good.

tests/integration/restricted/test.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
+
+set -ex
+
+RANDOM_CLUSTER_NAME_SUFFIX=`cat /dev/urandom | tr -cd 'a-f0-9' | head -c 7`
+
+CLUSTER_NAME="ds-test-restricted-${RANDOM_CLUSTER_NAME_SUFFIX}"
```
keegancsmith: may be useful to use a buildkite identifier if we have one? would be easier to link between the build and the cluster.

kubectl-apply-all.sh
efritz: It currently doesn't work anyway; https://github.com/kubernetes/kubernetes/issues/66430

uwedeportivo: i had no luck with apply -k --prune. it did worse and failed to create any objects

tests/integration/restricted/test.sh
uwedeportivo: ok, will do

tests/integration/restricted/test.sh
uwedeportivo: ok, will try

kubectl-apply-all.sh
ggilmore: I think this a rough problem. It doesn't just apply to services, but _any_ resource that's no longer referenced.

uwedeportivo:
we would like to salvage the prune functionality so this PR is in a holding pattern until we find a solution. prune is important because it relieves us from explicitly declaring migration steps with kubectl delete resources that are renamed/removed from new cluster versions. unfortunately prune is very buggy when used with -k.

kubectl-apply-all.sh
uwedeportivo: this was solved in a different PR

PR #571, opened 2020-03-27
Fix nonroot gitserver ssh key

DoomGerbil:
In the migration from root to non-root users in containers, this bit got missed.  Since `gitserver` no longer runs as root, the SSH key for repo syncs needs to be mounted for the proper user, not for root.

Missing this results in Sourcegraph silently failing to sync repos.  The only indicator of what's wrong is if you check the `gitserver-0` logs.  You'll see a ton of errors like this:

```
t=2020-03-27T23:14:16+0000 lvl=warn msg="error cloning repo" repo=github.com/improbable/inspector err="error cloning repo: repo github.com/improbable/inspector not cloneable: exit status 128 (output follows)\n\nHost key verification failed.\r\nfatal: Could not read from remote repository.\n\nPlease make sure you have the correct access rights\nand the repository exists.\n"
```

Once the key is emplaced in the right spot, everything works again. 👍

### DIFF
--- a/base/gitserver/gitserver.StatefulSet.yaml
+++ b/base/gitserver/gitserver.StatefulSet.yaml
@@ -50,7 +50,7 @@ spec:
           runAsGroup: 101
         # See the customization guide (../../../docs/configure.md) for information
         # about configuring gitserver to use an SSH key
-        # - mountPath: /root/.ssh
+        # - mountPath: /home/sourcegraph/.ssh
         #   name: ssh
       securityContext:
         fsGroup: 101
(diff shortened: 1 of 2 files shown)


beyang:
@uwedeportivo could you review this?

slimsag:
@uwedeportivo seems this could be relevant still, can you take a look?

DoomGerbil:
I can confirm that I needed to fix this again when updating to 3.15.1.

It might need to be done differently now, though, since running as non-root uses the non-root overlay.

uwedeportivo:
@DoomGerbil , i'm so sorry. don't know how i missed this PR.  as you noticed we decided in 3.15.1 to isolate overlays from base and keep base as root to minimize migration headaches and merging conflicts for customers that have not yet upgraded.  i took over your change and added it as https://github.com/sourcegraph/deploy-sourcegraph/pull/675. i'll close this PR but thanks again for submitting it. i promise i'll change my notifications and will react more timely next time :-)

PR #574, opened 2020-03-30
Rename lsif-server and split precise-code-intel services

efritz:
Split the lsif-server service into three service/deployment pairs:

- pci-api-server
- pci-worker
- pci-bundle-manager

Only the bundle manager has a persistent volume, which is distinct from the old lsif-server. We may need a manual migration step here for customers that want to keep their lsif data.

This mirrors the changes in https://github.com/sourcegraph/sourcegraph/pull/9376.


efritz:
Note: We need the actual image tags before it will pass CI.

docs/migrate.md
```diff
@@ -4,6 +4,19 @@ This document records manual migrations that are necessary to apply when upgradi
...
+
+### Existing installations: Migrating lsif-server
+
+Version 3.15 splits lsif-server into multiple services, but retains the same persistent volume. In order for the
+update to succeed, the old lsif-server service and deployment must be detached from the persistent volume. The
+following commands need to be run once after the updates have been applied to your cluster:
+
+```shell script
+kubectl delete svc lsif-server
+kubectl delete deployment lsif-server
```
slimsag: Maybe I am wrong, but I thought `kubectl-apply-all.sh` would automatically do this?

efritz: It doesn't seem to!

cc @uwedeportivo 

See https://github.com/kubernetes/kubernetes/issues/66430 and also https://github.com/sourcegraph/deploy-sourcegraph/pull/570 which uses -k instead of -f, which doesn't respect prune (apparently).

efritz: When running this on a test cluster the unreferenced services never went away (unless I missed a reference somewhere...?).

base/precise-code-intel/api-server.Deployment.yaml
```diff
@@ -0,0 +1,73 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    description: Precise code intelligence services for Sourcegraph
+  labels:
+    deploy: sourcegraph
+  name: pci-api-server
```
slimsag: I really dislike acronyms like `pci`. Such acronyms do not help anyone except those who already know what the thing is (same problem as LSIF).

I would be far happier with these being named `precise-code-intel-api-server`, `precise-code-intel-bundle-manager`, etc.

base/precise-code-intel/worker.Deployment.yaml
```diff
@@ -0,0 +1,73 @@
...
+        app: pci-worker
+    spec:
+      containers:
+      - env:
+        - name: PCI_NUM_APIS
+          value: '0'
+        - name: PCI_NUM_BUNDLE_MANAGERS
+          value: '0'
+        - name: PCI_NUM_WORKERS
+          value: '1'
```
slimsag: Can we have the default here be 2?

I would like us to generally use the same scaling mechanisms across kubernetes & docker-compose deployments where possible, and having this be 2 means we won't e.g. accidentally break the in-container replication without noticing. It'll also push us to scale workers using this in k8s deployments (instead of using replicas) which will inherently make us handle monitoring of these containers better.

docs/migrate.md
slimsag: Ah, got it, I must be wrong then. Seems reasonable to me, then.

base/precise-code-intel/api-server.Deployment.yaml
efritz: Should pci be replaced everywhere in https://github.com/sourcegraph/deploy-sourcegraph/pull/574/files as well?

efritz: Updated.

base/precise-code-intel/worker.Deployment.yaml
efritz: Both APIs and workers are up to 2.

docs/migrate.md
efritz: I'm not sure if this is affecting Pulumi in CI since we can't/aren't doing it in the automated tests.

efritz: https://buildkite.com/sourcegraph/deploy-sourcegraph/builds/2655#4134e558-9765-4e96-9a2c-4a5104fdbd9e

efritz: Oof ok nevermind. The reason that CI is failing is noted in a comment below.

efritz:
CI is currently failing because we have a 3.14 frontend and a brand-new set of precise-code-intel processes. If we start this on a test cluster you see that it gets stuck looping on waiting for migrations to run. Because these processes require a much higher schema version it never comes and they crash loop.

slimsag:
Well, it's good it is failing because running conflicting migration requirements like that would be bad.

This just means you need to bump the images in your PR here to match what precise-code-intel requires. i.e. if someone was to deploy this repository from `master`, what images would they need? we should merge those tags. This can be `insiders` with SHA digests pinned, for example.

efritz:
@slimsag CI is now passing (with all service builds updated to the `59615_2020-03-31_bf8119e` tag).

base/precise-code-intel/bundle-manager.Deployment.yaml
```diff
@@ -0,0 +1,82 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    description: Precise code intelligence services for Sourcegraph
```
slimsag, line 5:
```suggestion
    description: Stores and manages precise code intelligence bundles.
```

base/precise-code-intel/api-server.Deployment.yaml
```diff
@@ -2,17 +2,17 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   annotations:
-    description: HTTP server for LSIF-based precise code intelligence.
+    description: Precise code intelligence services for Sourcegraph
```
slimsag, line 5:
```suggestion
    description: Serves precise code intelligence requests.
```

base/precise-code-intel/worker.Deployment.yaml
```diff
@@ -0,0 +1,73 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    description: Precise code intelligence services for Sourcegraph
```
slimsag, line 5:
```suggestion
    description: Handles conversion of uploaded precise code intelligence bundles
```

docs/configure.md
```diff
@@ -471,7 +471,6 @@ If you want to specify a custom Redis server, you'll need specify the correspond
 
 - `sourcegraph-frontend`
 - `repo-updater`
-- `lsif-server`
```
slimsag, line 474: Does it no longer use Redis at all? Or do we need to update the instructions here?

efritz: It does not.

docs/migrate.md
```diff
@@ -4,6 +4,19 @@ This document records manual migrations that are necessary to apply when upgradi
 Sourcegraph versions. All manual migrations between the version you are upgrading from and the
 version you are upgrading to should be applied (unless otherwise noted).
 
+## 3.15
+
+### Existing installations: Migrating lsif-server
```
slimsag, line 9:
```suggestion
### Migrating lsif-server
```

I have no idea why the 3.14 section below says "Existing installations", because this entire document is _only_ about existing installations. Nobody would come here otherwise.

slimsag:
The more I keep looking at the migration required here the more I think it's not the right approach.

Most people are not using lsif-server today, and it seems odd to:

1. Introduce tech debt where we have the PVC named lsif-server just so (a few? one?) customer doesn't have to migrate or re-upload LSIF data.
2. Require all k8s admins to perform some manual migration steps (even simple manual migration steps make upgrading from e.g. 3.12 -> 3.15 very painful).

slimsag:
I would propose the following:

1. Rename the PVC to match precise-code-intel.
2. Document in the migrate.md file that lsif-server has been renamed, and admins will need to re-upload the LSIF data and basic code intelligence will be used in the meantime. Or, if that is not acceptable, to contact us for lossless migration instructions.

efritz:
@slimsag In this case how would we do that for dotcom?

slimsag:
Here is how you could make it look and write the migration instructions. All you would need to do is create a dumb e.g. `alpine` service deployment which mounts both PVCs and copies data from one to the other.:

---

To retain data from lsif-server, perform the following migration (which has some downtime for precise code intelligence):

1. Deploy 3.15, which creates a blank `precise-code-intel` PVC.
2. Take down the old `lsif-server` deployment, and the `precise-code-intel-bundle-manager` deployment so their PVCs are released:

```sh
kubectl delete svc lsif-server
kubectl delete deployment lsif-server
kubectl delete deployment precise-code-intel-bundle-manager
```

3. Create the deployment which will copy the old data from the `lsif-server` PVC to the new `precise-code-intel-bundle-manager` PVC (note: this may take a while):

```sh
kubectl apply -f docs/lsif-server-migrator.Deployment.yaml
```

4. Watch `kubectl logs lsif-server-migrator` until the copy completes (`Copy complete!`), then tear down that deployment and create the bundle manager again:

```sh
kubectl delete deployment lsif-server-migrator
# Whatver command to delete the old volume/PVC.
./kubectl-apply-all.sh
```

efritz:
@slimsag is `./configure` an ok place for this migration to live?

PR #576, opened 2020-03-31
fixes data dir file permissions for pgsql

uwedeportivo:
when i implemented migrate to non-root user i missed one thing:

i tested migrating, tested rollback, tested fresh. what i didn't test: a restart of pgsql after the migration.

according to https://github.com/kubernetes/kubernetes/issues/2630#issuecomment-70041443 volumes are mounted with 770 on gke in this situation.

the other deployments and stateful sets in our installation don't care about the 770 file permissions, so they keep working. but pgsql is insistent that it's 700 or 750.


overlays/migrate-to-nonroot/kustomization.yaml
```diff
@@ -7,7 +7,6 @@ patchesStrategicMerge:
   - grafana/grafana.Deployment.yaml
   - indexed-search/indexed-search.StatefulSet.yaml
   - lsif-server/lsif-server.Deployment.yaml
-  - pgsql/pgsql.Deployment.yaml
```
efritz: This is still safe? We can run the init container regardless if this kustomization being run?

uwedeportivo: that initContainer is what the deleted kustomization did

PR #577, opened 2020-03-31
non-root user migration: add /mnt/cache volumes from base

uwedeportivo:
to kustomization

they are emptyDir volumes which could be affected if pod stays on same node during migration to non-root user and then cannot read/write from the cache

this does not address configure/ssd /mnt/cache volumes.

i will change the pod-tmp-gc daemonset there probably

### DIFF
--- a/overlays/migrate-to-nonroot/frontend/sourcegraph-frontend.Deployment.yaml
+++ b/overlays/migrate-to-nonroot/frontend/sourcegraph-frontend.Deployment.yaml
@@ -1,0 +1,16 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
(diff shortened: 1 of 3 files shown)


overlays/migrate-to-nonroot/frontend/sourcegraph-frontend.Deployment.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: sourcegraph-frontend
+spec:
+  template:
+    spec:
+      initContainers:
```
efritz: just going all out on init containers huh?

Issue #572, opened 2020-03-27
SSD cache permissions


efritz:
After running the nonroot overlay on DotCom, we found that the frontend pods (among others) that amount a cache directory in a shared SSH do not have the correct permissions for writing.

https://github.com/sourcegraph/deploy-sourcegraph/tree/master/configure/ssd

These services mount the SSD cache:

- sourcegraph-frontend
- searcher
- symbols
- lang-go
- lang-typescript
- pod-tmp-gc

efritz:
This was resolved on DotCom via

```
gcloud compute instances list | grep dot-com
```

and

```
gcloud compute ssh gke-dot-com-dot-com-{node-id} --command 'sudo chmod -R 0777 /mnt/disks/ssd0/pod-tmp'
```

uwedeportivo:
https://github.com/sourcegraph/deploy-sourcegraph/pull/583 fixes this

PR #583, opened 2020-04-02
only use fsGroup at cluster creation

uwedeportivo:
This kustomization injects a fsGroup security context in each pod so that the volumes are mounted with the
specified supplemental group id and non-root pod users can write to the mounted volumes.

This is only done once at cluster creation time so this overlay is only referenced by the `create-new-cluster.sh`
script.

The reason for this approach is the behavior of fsGroup: on every mount it recursively chmod/chown the disk to add
the group specified by fsGroup and to change permissions to 775 (so group can write). This can take a long time for
large disks and sometimes times out the whole pod scheduling.

If we only do it at cluster creation time (when the disks are empty) it is fast and since the disks are persistent
volumes we know that the pod user can write to it even without the fsGroup and subsequent apply operations.

In Kubernetes 1.18 fsGroup gets an additional [feature](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods)
called `fsGroupChangePolicy` that will allow us to control the chmod/chown better.


uwedeportivo:
fixes https://github.com/sourcegraph/deploy-sourcegraph/issues/572

uwedeportivo:
both migration and fresh were tested on GKE test cluster

integration test broken because i'm waiting for new buildkite agent with kustomize in it

.buildkite/pipeline.yml
```diff
@@ -5,8 +5,8 @@ steps:
     command: .buildkite/verify-label.sh
   - label: ":k8s:"
     command: .buildkite/verify-rbac-labels.sh
-  - label: ':lipstick:'
-    command: .buildkite/prettier-check.sh
+  #- label: ':lipstick:'
+  #  command: .buildkite/prettier-check.sh
```
keegancsmith, line 9: why is this commented out?

uwedeportivo: temporarily, because yarn was giving me trouble too (on top of atlassian package manager and other stuff). yesterday the software goblins won

DoomGerbil:
FYI @uwedeportivo , this change broke Improbable's Sourcegraph 3.14 deployment, because we mount an SSH secret into gitserver.

This secret data needs to be readable by the now-non-root user, and this change makes that no longer true (with `fsGroup` unset, the key has `root:root -rw-------` perms, and with it, the perms are `root:sourcegraph -rw-r-----` which means the user can access the key).  The result is that gitserver is unable to sync changes from repos because the sourcegraph user is unable to read the `ssh` secret which is mounted owned by `root`.

I've re-added this to our internal deployment, but this is the second SSH fetch/clone breakage recently (see https://github.com/sourcegraph/deploy-sourcegraph/pull/571 for a PR I raised to fix the other one).

beyang:
@DoomGerbil Sorry you had to deal with that and thank you for submitting #571. I've filed https://github.com/sourcegraph/sourcegraph/issues/9722 to discuss how we can catch this in the future before it hits you.

PR #586, opened 2020-04-03
kustomize: delete kustomization.yaml

uwedeportivo:
it triggers a (harmless but annoying) error message and exit with non-zero code when doing

```
kubectl apply -f base --recursive
```

this breaks CI and confuses customers


uwedeportivo:
superseded by https://github.com/sourcegraph/deploy-sourcegraph/pull/589

PR #587, opened 2020-04-03
make memory units consistent

beyang:


beyang:
> This is good, but just FYI 2G != 2Gi (even more of a reason to standardize)

Yep, that's the reason for standardizing. A customer bumped the numerical value for memory limits and memory requests to the same number, didn't change the units, which caused an error, because 2Gi > 2G (and thus mem request exceeded mem limit).

PR #588, opened 2020-04-03
move base to subdir to avoid kustomization.yaml

uwedeportivo:
avoids the complications with kustomization.yaml by moving the resource files into subdir and referencing them directly when doing kubectl apply -f


uwedeportivo:
https://gravitational.com/blog/kubernetes-kustomize-kep-kerfuffle/

uwedeportivo:
https://github.com/kubernetes-sigs/kustomize/issues/738#issuecomment-458699417

beyang:
Will this cause any merge conflicts with forks that have made changes to files in `base`? (I.e., is git smart enough to deal with renames in this case?)

uwedeportivo:
superseded by https://github.com/sourcegraph/deploy-sourcegraph/pull/589

PR #594, opened 2020-04-06
build author label escaping

uwedeportivo:
BUILDKITE_CREATOR_EMAIL is not good because it can have too many characters that are not allowed
(it's debatable if names are worse actually) and also some github emails are redirect emails to protect contributor privacy


uwedeportivo:
the switch to ingress is needed for https://github.com/sourcegraph/deploy-sourcegraph/pull/559

uwedeportivo:
better fixes were landed in https://github.com/sourcegraph/deploy-sourcegraph/pull/559

Issue #387, opened 2019-10-02
Grafana persistent volume size


aisbaa:
Hi, I'm just curious why grafana persistent volume is so big 200Gi, which is the same size as prometheus. To my understanding grafana does not store time series data on disk, it only needs to store dashboards and datasource configuration, sometimes user passwords. None of these should required a lot of space, but maybe I'm missing something.

Here's the file https://github.com/sourcegraph/deploy-sourcegraph/blob/master/base/grafana/grafana.PersistentVolumeClaim.yaml

Some examples online provide dashboard configuration in ConfigMap and do not assign any persistent volume to grafana pod. I think sharing dashboard configuration in ConfigMap is One of the best ways to do that!

beyang:
@aisbaa you are absolutely right (and sorry about the delay here; we are migrating issues in this repository to the Sourcegraph main issue tracker to better keep track of them). Feel free to reduce the disk size. We'll also update this in our next version.

aisbaa:
Thats so cool, thank you.

PR #616, opened 2020-04-10
reduce grafana pvc from 200Gi to 2Gi

beyang:
Grafana disk usage is only 60mb on k8s.sgdev.org and sourcegraph.com, so why do we ask for 200gb? Fixes https://github.com/sourcegraph/deploy-sourcegraph/issues/387.

### DIFF
--- a/base/grafana/grafana.PersistentVolumeClaim.yaml
+++ b/base/grafana/grafana.PersistentVolumeClaim.yaml
@@ -9,5 +9,5 @@ spec:
   - ReadWriteOnce
   resources:
     requests:
-      storage: 1Gi
+      storage: 2Gi
   storageClassName: sourcegraph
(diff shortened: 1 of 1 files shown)


slimsag:
I think 2Gi is a safer bet; Grafana _does_ do caching of metrics on dashboards and as we start to add many more this will add up.

base/grafana/grafana.PersistentVolumeClaim.yaml
```diff
@@ -9,5 +9,5 @@ spec:
   - ReadWriteOnce
   resources:
     requests:
-      storage: 200Gi
+      storage: 1Gi
```
beyang, line 12:
```suggestion
      storage: 2Gi
```

PR #621, opened 2020-04-14
jaeger-all-in-one: non-root permissions on /tmp

beyang:
This addresses https://github.com/sourcegraph/deploy-sourcegraph/pull/559#discussion_r405182789

Post-merge:
- [ ] Move Docker images to https://github.com/sourcegraph/sourcegraph/tree/master/docker-images

### DIFF
--- a/base/jaeger/jaeger.Deployment.yaml
+++ b/base/jaeger/jaeger.Deployment.yaml
@@ -27,7 +27,7 @@ spec:
         prometheus.io/port: "16686"
     spec:
         containers:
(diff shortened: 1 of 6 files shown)


slimsag:
Note: when you move this over, you should use an almost identical build.sh as https://github.com/sourcegraph/sourcegraph/blob/master/docker-images/grafana/build.sh

This will ensure the image gets built on CI automatically, as well as that it gets versioned alongside Sourcegraph (`jaeger-all-in-one:3.15.0` / `jaeger-all-in-one:insiders` NOT `jaeger-all-in-one:1.17.1`).

You should also make sure your Dockerfile has these labels: https://github.com/sourcegraph/sourcegraph/blob/master/docker-images/prometheus/Dockerfile#L20-L30

PR #638, opened 2020-04-17
guarantee Grafana and Prometheus resources

slimsag:
See inline comments for details.

This change should be cherry-pick'd onto 3.15 before release.

### DIFF
--- a/base/grafana/grafana.Deployment.yaml
+++ b/base/grafana/grafana.Deployment.yaml
@@ -35,13 +35,17 @@ spec:
           name: data
         - mountPath: /sg_config_grafana/provisioning/datasources
           name: config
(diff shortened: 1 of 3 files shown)


beyang:
Could we keep the requests low (while raising the limits)?

If we do raise the requests, could you add this to the CHANGELOG, in case this means some customers have to expand the size of their cluster.

Otherwise LGTM.

slimsag:
> Could we keep the requests low (while raising the limits)?

No, this is the whole point of the change: If the requests are low, it means another service running on the same pod and consuming too much memory can conflict and cause Prometheus/Grafana to OOM. Thus, you end up with a service OOMing and the things supposed to be monitoring it also OOMing because of it

PR #656, opened 2020-04-23
back to root user

uwedeportivo:
this brings the default base back to root user and provides an overlay for customers that want to run as non-root.

kubectl-apply-all.sh is back to its previous glory along with base.

overlays are self-contained but do need the overlay-generate-cluster.sh to operate.

overlay-generate-cluster.sh takes two arguments: the overlay to use and the directory where it outputs the generated cluster.

one additional advantage of this approach is that overlay users can now also do apply -f --prune on the generated manifests. before they couldn't use prune with apply -k (well, they could but would run the risk that their whole deployment would disappear).

i will change documentation to describe all this if this approach is approved.


slimsag:
briefly adding some additional context for e.g. Keegan or others who weren't in the distribution call:

In short, a customer upgraded to this but the experience with merge conflicts etc. was quite painful so we're now trying to undo it by-default and make it optional as a kustomize overlay so that most of our users do not have to go through the pain of migrating to this non-root world.

PR #664, opened 2020-04-28
cherry-pick back to root

uwedeportivo:
tough one


uwedeportivo:
tested several permutations with test cluster on gke

uwedeportivo:
with my apologies and in the interest of shipping the security fix, i'll land this

PR #665, opened 2020-04-29
update docker image tags to 3.14.4

uwedeportivo:
don't wait for renovate


uwedeportivo:
with my apologies and in the interest of shipping the security fix, i'll land this

PR #674, opened 2020-05-04
move from deployment to statefulset

davejrt:
a template for the migration to statefulsets where a deployment has a single replicate with a volume attached see issue sourcegraph/sourcegraph#2076


davejrt:
@sourcegraph/distribution raising this as a lifting off point to discuss what migrating to statefulsets would look like to the rest of the services.

uwedeportivo:
i read through the issue. i'm not sure i understand. our pvc says "ReadWriteOnce", isn't that enough to prevent corruption ? the only loophole i can think of is if a rolling update puts the new pod on the same node as the old pod. maybe that is a possibility. on the other hand moving to a statefulset is harmless, so i guess you're right: we should do it for singletons too.

davejrt:
It's harmless and I'd say whilst the accessmode does that, and we aren't specifying anti affinity rules this is the best way to avoid it, especially for those customers who could run into trouble and don't have the immediate knowledge on hand to avoid or fix this issue.

davejrt:
As mentioned, this is more of an example just to ensure we're happy with the approach before I proceed to change the rest of the services.

beyang:
LGTM. One thing to note is that for Grafana is that I don't think we really care about the data on the PV, because the dashboards all come from the ConfigMap. The site admin can in theory create their own in the GUI, and these *would* be lost in this migration, but I don't know of any customers that have actually done this (@slimsag should we actively discourage this in the docs?).

We *do*, however, care about the data for other services (e.g., Postgres, Redis), and the overhead of doing the migration might mean we push off migrating those to StatefulSets, unless we can think of a clean way to do the migration for all customers.

davejrt:
Sounds good. I am going to put a separate PR with something like redis with recreate as an alternative for the team to review/discuss.

PR #675, opened 2020-05-04
non-root: document mount path for ssh config

uwedeportivo:
Documents where to mount ssh config for gitserver in non-root overlay.

See https://github.com/sourcegraph/deploy-sourcegraph/pull/571 also

### DIFF
--- a/overlays/non-root/gitserver/gitserver.StatefulSet.yaml
+++ b/overlays/non-root/gitserver/gitserver.StatefulSet.yaml
@@ -12,9 +12,20 @@ spec:
           allowPrivilegeEscalation: false
           runAsUser: 100
           runAsGroup: 101
+      # See the customization guide (../../../docs/configure.md) for information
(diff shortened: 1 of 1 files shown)


overlays/non-root/gitserver/gitserver.StatefulSet.yaml
```diff
@@ -12,9 +12,20 @@ spec:
...
+      # the volumes section below also.
+      # - mountPath: /home/sourcegraph/.ssh
+      #   name: ssh
       - name: jaeger-agent
         securityContext:
           # Required to prevent escalations to root.
           allowPrivilegeEscalation: false
           runAsUser: 100
           runAsGroup: 101
+      # volumes:
```
beyang, line 27:
```suggestion
      # Uncomment this section along with the other section (see the docstring above the `mountPath` block above) if you're using a SSH key for git cloning.
      # volumes:
```

PR #676, opened 2020-05-05
migrating update strategy to recreate

davejrt:
addresses sourcegraph/sourcegraph#2076 as an example of what can be applied to other services.

Assuming customers have deployed this using `kubectl apply` and not `kubectl create` no other scripting should be required to change this property on the deployment.

### DIFF
--- a/base/pgsql/pgsql.Deployment.yaml
+++ b/base/pgsql/pgsql.Deployment.yaml
@@ -14,10 +14,7 @@ spec:
     matchLabels:
       app: pgsql
   strategy:
-    rollingUpdate:
-      maxSurge: 1
-      maxUnavailable: 1
-    type: RollingUpdate
+    type: Recreate
   template:
     metadata:
       labels:
--- a/base/precise-code-intel/bundle-manager.Deployment.yaml
+++ b/base/precise-code-intel/bundle-manager.Deployment.yaml
@@ -14,10 +14,7 @@ spec:
     matchLabels:
       app: precise-code-intel-bundle-manager
   strategy:
-    rollingUpdate:
-      maxSurge: 1
-      maxUnavailable: 1
-    type: RollingUpdate
+    type: Recreate
   template:
     metadata:
       labels:
--- a/base/prometheus/prometheus.Deployment.yaml
+++ b/base/prometheus/prometheus.Deployment.yaml
@@ -14,10 +14,7 @@ spec:
     matchLabels:
       app: prometheus
   strategy:
-    rollingUpdate:
-      maxSurge: 1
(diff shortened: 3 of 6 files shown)


uwedeportivo:
this basically avoids having old and new pod on the same node trying to access same persistent volume, correct ? it kills the old one before bringing up the new one, right ?

davejrt:
> this basically avoids having old and new pod on the same node trying to access same persistent volume, correct ? it kills the old one before bringing up the new one, right ?

Exactly this @uwedeportivo

keegancsmith:
So the reason we haven't done this change before is you can't just apply this change to deployment resources. The rollout strategy is an immutable field, so it requires recreating the deployment. IE we would need to introduce some documentation/script around automating this rollout.

I think another thing to take into account is instead of making these deployments, it is probably more appropriate to use StatefulSets for these services. In theory it would give us a path forward to make them HA.

davejrt:
My local testing indicates you can now update the strategy field via edit and apply:

```
davet@dave-surface-2:~/scratch $ kubectl get deploy mysql -o jsonpath="{.spec.strategy}"
map[rollingUpdate:map[maxSurge:1 maxUnavailable:1] type:RollingUpdate]

davet@dave-surface-2:~/scratch $ kubectl apply -f mysql-deployment-recreate.yaml
service/mysql unchanged
deployment.apps/mysql configured
davet@dave-surface-2:~/scratch $ kubectl get deploy mysql -o jsonpath="{.spec.strategy}"
map[type:Recreate]
davet@dave-surface-2:~/scratch $
```

This isn't to say I don't disagree that for stateful applications we should be using statefulsets, a change like this addresses the immediate concern

keegancsmith:
woah cool! A nice rare change to k8s which improves a pain point rather than introducing a new abstraction :P

davejrt:
I found similar documentation, in that there is nothing concrete just changes that allude to it. Testing as far back as 1.11.10 in minikube seems to work. 

Is the changelog the best place to document this, or an alternative doc specific to an upgrade from whatever release these changes are in?

keegancsmith:
> Is the changelog the best place to document this, or an alternative
> doc specific to an upgrade from whatever release these changes are in?

Changelog is good. We also have a migrations.md in this repo which may
be appropriate (most admins check with that before upgrading).

davejrt:
I've updated the rest of deployments in the PR to ensure they all make it into the same release. Please feel free to re-review. Migrate.md has been updated also

davejrt:
> LGTM. I just took a look at prod and noticed one other deployment which has a volume mounted: deploy/precise-code-intel-bundle-manager.

Fixed

PR #678, opened 2020-05-06
overlays: readme fixes and namespaced prometheus rbac

uwedeportivo:
Clean up in README instructions and patched a `namespaced` overlay omission (the namespace value needs to be set in the cluster rolebinding for prometheus)

### DIFF
--- a/overlays/non-privileged/README.md
+++ b/overlays/non-privileged/README.md
@@ -13,5 +13,5 @@ To use it, execute the following command from the root directory of this reposit
 After executing the script you can apply the generated manifests from the `generated-cluster` directory:
 
 ```shell script
-kubectl apply --prune -l deploy=sourcegraph -f generated-cluster --recursive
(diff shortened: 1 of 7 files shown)


overlays/non-privileged/README.md
```diff
@@ -13,5 +13,5 @@ To use it, execute the following command from the root directory of this reposit
 After executing the script you can apply the generated manifests from the `generated-cluster` directory:
 
 ```shell script
-kubectl apply --prune -l deploy=sourcegraph -f generated-cluster --recursive
```
beyang, line 16: Will this cause issues on update for any customers who previously omitted the namespace? If so, how will we address that?

daxmc99:
Whoops, did not mean to close

overlays/non-privileged/README.md
uwedeportivo: no issue. it works without. i just wanted to make it explicit that they are working in a namespace in case they use kubectl to do log/describe and other commands. doing apply works because the namespace is declared in the manifests.

PR #684, opened 2020-05-11
move some docs to docs.sourcegraph.com

uwedeportivo:


uwedeportivo:
with apologies, i'm going to land this so i don't have to play catchup

PR #685, opened 2020-05-12
apply-all script: allow specifying additional flags

uwedeportivo:
This PR allows passing additional flags into the kubectl command used by kubectl-apply-all.

Can be used to 

```
./kubectl-apply-all.sh --dry-run
```

```
./kubectl-apply-all.sh --dry-run --validate
```

or 

```
./kubectl-apply-all.sh  -l sourcegraph-resource-requires=no-cluster-admin
./kubectl-apply-all.sh  -l sourcegraph-resource-requires=cluster-admin
```

to separate manifests for cluster-wide and non-cluster-wide admins (addresses issue https://github.com/sourcegraph/sourcegraph/issues/9723, modulo documenting it).

Kubernetes-specific change with no docker counterpart.

### DIFF
--- a/base/backend.Service.yaml
+++ b/base/backend.Service.yaml
@@ -6,7 +6,7 @@ metadata:
       the same node if possible.
   labels:
     deploy: sourcegraph
-    rbac-admin: non-escalated
+    sourcegraph-rbac-admin: non-escalated
     group: backend
   name: backend
 spec:
--- a/base/backend.Service.yaml
+++ b/base/backend.Service.yaml
@@ -6,6 +6,7 @@ metadata:
       the same node if possible.
   labels:
     deploy: sourcegraph
+    rbac-admin: non-escalated
     group: backend
   name: backend
 spec:
--- a/base/backend.Service.yaml
+++ b/base/backend.Service.yaml
@@ -6,7 +6,7 @@ metadata:
       the same node if possible.
   labels:
     deploy: sourcegraph
-    sourcegraph-rbac-admin: non-escalated
+    sourcegraph-resource-requires: no-cluster-admin
     group: backend
   name: backend
 spec:
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ generated/
 .pulumi
 Pulumi.*.yaml
 /pkg/gen/openapi-specs
+generated-cluster/
--- a/base/frontend/sourcegraph-frontend-internal.Service.yaml
+++ b/base/frontend/sourcegraph-frontend-internal.Service.yaml
@@ -4,7 +4,7 @@ metadata:
   labels:
     app: sourcegraph-frontend
     deploy: sourcegraph
-    sourcegraph-rbac-admin: non-escalated
+    sourcegraph-resource-requires: no-cluster-admin
   name: sourcegraph-frontend-internal
 spec:
   ports:
--- a/base/frontend/sourcegraph-frontend-internal.Service.yaml
+++ b/base/frontend/sourcegraph-frontend-internal.Service.yaml
@@ -4,7 +4,7 @@ metadata:
   labels:
     app: sourcegraph-frontend
     deploy: sourcegraph
-    rbac-admin: non-escalated
+    sourcegraph-rbac-admin: non-escalated
   name: sourcegraph-frontend-internal
 spec:
   ports:
--- a/base/frontend/sourcegraph-frontend-internal.Service.yaml
+++ b/base/frontend/sourcegraph-frontend-internal.Service.yaml
@@ -4,6 +4,7 @@ metadata:
   labels:
     app: sourcegraph-frontend
     deploy: sourcegraph
+    rbac-admin: non-escalated
   name: sourcegraph-frontend-internal
 spec:
   ports:
--- a/base/frontend/sourcegraph-frontend.Deployment.yaml
+++ b/base/frontend/sourcegraph-frontend.Deployment.yaml
@@ -5,7 +5,7 @@ metadata:
     description: Serves the frontend of Sourcegraph via HTTP(S).
   labels:
     deploy: sourcegraph
-    sourcegraph-rbac-admin: non-escalated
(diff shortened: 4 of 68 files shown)


uwedeportivo:
i had to add a label value `rbac-admin=non-escalated` to all manifests that are not labeled `rbac-admin=escalated` because `kubectl apply -f --prune` does funky stuff if you give it a label condition `-l rbac-admin!=escalated` (it tries to delete kube-system)

uwedeportivo:
tested the networkpolicy on a test cluster. made sure it denies traffic from another namespace

base/backend.Service.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
       the same node if possible.
   labels:
     deploy: sourcegraph
+    rbac-admin: non-escalated
```
slimsag, line 9: Should we use a different label like `sourcegraph-rbac-admin` as @daxmc99 noted would've helped a customer understand it more easily to be a label of ours and not one that e.g. Kubernetes interprets?

uwedeportivo: i don't mind doing it. the problem is the label already exists

slimsag: By problem, do you mean we would have to rename it in files not modified here currently and you'd prefer doing it in a separate PR -- or that it introduces a problem because someone may be relying on it currently or something else?

uwedeportivo: the latter. it has been there for a while, it's mentioned in our docs and readme. not sure if somebody relies on it

slimsag: Doesn't seem that important to me. We just need to mention it in the migrate.md docs that it has been renamed, so if you use it (most do not) you'll need to use the new name.

uwedeportivo: ok, will do

uwedeportivo: the deploy label stays the same, i assume ?

slimsag: Yeah, I would think so

base/backend.Service.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
       the same node if possible.
   labels:
     deploy: sourcegraph
+    sourcegraph-rbac-admin: non-escalated
```
slimsag, line 9: Sorry to put this in last-minute: should we reconsider the name and values too? As I understand it they are:

- `sourcegraph-rbac-admin: escalated`
- `sourcegraph-rbac-admin: non-escalated`

But.. what do these really mean? What is "escalated role based access control administration"?

slimsag: In other words, I wonder what we're describing here in practice and whether or not it would make sense to me when explained with docs and someone more experienced with k8s like @daxmc99 without docs.

uwedeportivo: no problem, it's just search/replace :-)

there's two subsets of admins apparently: cluster-wide admins have permission to create/modify/delete cluster-wide resources (like ClusterRoles and ClusterRoleBindings). sometimes only these people can create a storage class for example. this group are the admins with super powers.

the other group are constrained to a namespace and can only create resources in that namespace.

this label marks things as needing the first group to create them or not. unfortunately you cannot mix a label inequality with -prune because all sorts of hell will break loose. so you have to name the non-special resources as not needing admin super powers and select them also with a label equality clause.

of course you can define even more granular and specific permissions with roles, but this is the most common.

uwedeportivo: you can then ask the ones with super powers to do
```
./kubectl-apply-all.sh  -l sourcegraph-rbac-admin=escalated
```

and the other group can take over and do the

```
./kubectl-apply-all.sh  -l sourcegraph-rbac-admin=non-escalated
```

uwedeportivo: so any name/value change is welcome (now that you know what they mean) :-)

slimsag: Got it, that is a very helpful explanation, thank you! In this case, I might propose:

```
sourcegraph-service-requires=admin
sourcegraph-service-requires=non-admin
````

This makes sense to me as an outsider to the norms of k8s, but does it make sense to you @uwedeportivo and others?

Comment (uwedeportivo) [reply base/backend.Service.yaml:9]: it's not just service

how about

```
sourcegraph-resource-requires=cluster-admin
sourcegraph-resource-requires=no-cluster-admin
````

Comment (slimsag) [reply base/backend.Service.yaml:9]: That looks great to me!

Comment (slimsag) [reply base/backend.Service.yaml:9]: I'd like to have buy-in from 1 other on the team that this makes sense to them (Geoffrey, Dax, Dave, Beyang, etc)

Comment (uwedeportivo) [reply base/backend.Service.yaml:9]: sure. i'll wait

Comment (keegancsmith) [reply base/backend.Service.yaml:9]: Not to jump on the bandwagon, but are these more appropriate as annotations instead of labels? Do we imagine k8s tooling to need to automatically select on these labels?

Comment (uwedeportivo) [reply base/backend.Service.yaml:9]: yes, we do. we want to

```
./kubectl-apply-all.sh  -l sourcegraph-rbac-admin=escalated
```
and 
```
./kubectl-apply-all.sh  -l sourcegraph-rbac-admin=non-escalated
```

uwedeportivo:
companion PR in the docs:

https://github.com/sourcegraph/sourcegraph/pull/10583/files#diff-2c5e1493332a3c7d2432ee2891114855R558 explains NetworkPolicy

https://github.com/sourcegraph/sourcegraph/pull/10583/files#diff-8451e3fd559414916e3f7e378f26ce07R84 explains cluster-admin privileges

base/backend.Service.yaml
davejrt: +1 from me. This makes good sense delegating responsibilities between those with elevated admin privileges and those without.

daxmc99: I'm a fan of the prefix. I think it helps communicate that these are specifically used by Sourcegraph.

slimsag:
🎉

PR #686, opened 2020-05-13
docs: update k8s.eks

bobheadxi:
closes https://github.com/sourcegraph/sourcegraph/issues/9724

https://github.com/sourcegraph/deploy-sourcegraph/blob/docs/aws-eks/docs/k8s.md

Some comments (mostly based on https://github.com/sourcegraph/sourcegraph/issues/9724):
* as pointed out in the ticket it seems that the EKS getting-started guide for nodegroups differs from the nodegroup-specific documentation (https://docs.aws.amazon.com/eks/latest/userguide/worker_node_IAM_role.html), but the docs here follow the getting started guide, so it should be safe to stick with?
* updated the format of the `k8s.md` table to hopefully make it more clear that some of those links point to sourcegraph-specific setup
* fixed a number of outdated links

I ran it through the Sourcegraph install steps too - first time around, I configured `t2.medium`'s and the sourcegraph deploy kept blowing up 😢 using the recommended cluster size settings got it up and running though. Overall, not too painful of an experience, but there are quite a few steps

image image

Sister [deploy-sourcegraph-docker](https://github.com/sourcegraph/deploy-sourcegraph-docker) change: n/a (I think? since this is just a docs change)

*note: doing this because it's the first issue listed in my project planning document 😄*


slimsag:
> Sister deploy-sourcegraph-docker change: n/a (I think? since this is just a docs change)

Yep, that's right! :)

davejrt:
LGTM

PR #687, opened 2020-05-14
test validates instance

uwedeportivo:
adds a validation step to a test run that also uses gitserver ssh keys

implements https://github.com/sourcegraph/sourcegraph/issues/9722

companion PR https://github.com/sourcegraph/src-cli/pull/200


tests/integration/restricted/test.sh
```diff
@@ -76,3 +89,6 @@ kubectl -n ns-sourcegraph port-forward svc/sourcegraph-frontend 30080 &
 CLEANUP="kill $!; $CLEANUP"
 sleep 2 # (initial delay in port-forward activating)
 curl --retry-connrefused --retry 2 --retry-delay 10 -m 30 http://localhost:30080
+
+# run a validation script against it
+src-alpha -endpoint http://localhost:30080 validate -context github_token=$GITHUB_TOKEN validate.star
```
bobheadxi, line 94: It seems the plan is to remove `-endpoint`: https://github.com/sourcegraph/src-cli/pull/235

tests/integration/restricted/validate.star
```diff
@@ -0,0 +1,27 @@
+def validate():
+    if cloned:
+       count = src_search_match_count("repo:^github.com/sourcegraph-testing/repo-ssh-keys-test$ gallery-director-drone-vacant-blizzard count:99999")
+       return count > 0
+    else:
+       return False
+
+src_create_first_admin("e2e@sourcegrah.com", "e2e-test-user", "123123123-e2e-test")
```
bobheadxi, line 8: I'm guessing these are utility functions provided by https://github.com/sourcegraph/src-cli/pull/200 right? Judging from the names this validation looks good

Probably a better question for https://github.com/sourcegraph/src-cli/pull/200 but what happens if any of these step fails? Do these test accounts and repos get removed?

tests/integration/restricted/test.sh
uwedeportivo: thanks for pointing that out

tests/integration/restricted/validate.star
uwedeportivo: no, they won't get removed. it's sort of a "who's watching the watcher" situation. if they fail then i don't have much of a chance of cleaning either, but you're right: i should at least put some code in there to try

bobheadxi: For other cases, might be neat to provide a `defer`-style thing for cleanup purposes

uwedeportivo:
ok, switched to json spec for validation, give it another look

PR #707, opened 2020-05-22
add cadvisor to deployment

bobheadxi:
closes https://github.com/sourcegraph/sourcegraph/issues/9791

I'm not entirely familiar with this so... it's mostly a stab in the dark 😨 

* add cAdvisor to `base/`, based on https://github.com/google/cadvisor/tree/master/deploy/kubernetes and some copy-pasta from the other `base/` configurations
* ~kustomize [says no to removing things in overlays](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/eschewedFeatures.md#removal-directives), so I've added a step to `overlay-generate-cluster.sh` to strip out cAdvisor things for relevant overlays (currently `non-privileged*`)~ => https://github.com/sourcegraph/deploy-sourcegraph/pull/707#discussion_r429461181
* update prometheus to correctly pick up on pods that want to be scraped
* expose cadvisor on 48080 on each node

Sister [deploy-sourcegraph-docker](https://github.com/sourcegraph/deploy-sourcegraph-docker) change: n/a, docker-compose [already has cadvisor](https://github.com/sourcegraph/deploy-sourcegraph-docker/blob/master/docker-compose/docker-compose.yaml#L467)


base/cadvisor/README.md
```diff
@@ -0,0 +1,5 @@
+# cAdvisor
+
+[cAdvisor](https://github.com/google/cadvisor) provides container users an understanding of the resource usage and performance characteristics of their running containers. It is a running daemon that collects, aggregates, processes, and exports information about running containers.
+
+cAdvisor is part of the default Sourcegraph cluster installation, and deployed as a [Kubernetes DaemonSet](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/). This setup is based on the [official cAdvisor Kubernetes Daemonset configuration](https://github.com/google/cadvisor/tree/master/deploy/kubernetes).
```
slimsag, line 5: We should add a short paragraph below here to the effect of "we use our own cadvisor deployment not the metrics k8s exports itself because it is often outdated" or something like that

base/cadvisor/cadvisor.DaemonSet.yaml
```diff
@@ -0,0 +1,68 @@
...
+      name: cadvisor
+  template:
+    metadata:
+      labels:
+        name: cadvisor
+    spec:
+      serviceAccountName: cadvisor
+      containers:
+      - name: cadvisor
+        image: k8s.gcr.io/cadvisor:v0.30.2
```
slimsag, line 23: All images here need to be in the form of `index.docker.io/sourcegraph/...:<sourcegraph_version>`

you just need to republish this image, it's really simple just e.g. copy what is done in this directory here: https://github.com/sourcegraph/sourcegraph/blob/master/docker-images/syntax-highlighter/build.sh

You should have a `docker-images/cadvisor/build.sh` and then once that gets merged this line should become:

```
index.docker.io/sourcegraph/cadvisor:insiders@sha256:...
```

And once we do a real release with it, it'll become:

```
index.docker.io/sourcegraph/cadvisor:3.17.0@sha256:...
```

so it's easy to know "is this the right image for my sourcegraph version"

slimsag: Once you do this, can you also update the image in the deploy-sourcegraph-docker repo? we should be doing this there too, must've been missed by accident

slimsag:
very nice job taking a stab at this!

@uwedeportivo could you review / give advice on the overlay approach here? Would setting `replicas: 0` be better or worse, I wonder, too?

uwedeportivo:
will look in a bit

base/cadvisor/README.md
bobheadxi: added a paragraph in 867e1582214a838bfba07f0ae9625fe716cb76da !

base/cadvisor/cadvisor.DaemonSet.yaml
slimsag: No, you literally want a file like this in our main repo `docker-images/cadvisor/build.sh`:

```sh
#!/usr/bin/env bash
cd "$(dirname "${BASH_SOURCE[0]}")"
set -ex

# This merely re-tags the image to match our official versioning scheme.
docker pull k8s.gcr.io/cadvisor:v0.30.2@sha256:SOMETHING
docker tag k8s.gcr.io/cadvisor:v0.30.2@sha256:SOMETHING "$IMAGE"
```

you just gotta find the SHA + send that PR to the main repo and you're good

base/cadvisor/README.md
slimsag: I think GitHub is breaking? it doesn't show this commit in your PR :/

base/cadvisor/cadvisor.DaemonSet.yaml
bobheadxi: ab gotcha, thanks! the example you linked for `syntax-highlighter` seemed to indicate the image being tagged is built in `infrastructure`

base/cadvisor/README.md
bobheadxi: I think it's pushed, should show up eventually 😅

base/cadvisor/cadvisor.DaemonSet.yaml
bobheadxi: `deploy-sourcegraph` is at v0.33, latest is v0.36 - should I just bump to the newest one?

slimsag: Yes please update us to the latest version in all locations, should "just work"

slimsag: the example I linked was just the simplest retagging example I could find, it being built in infra is unrelated to the problem here :)

base/cadvisor/cadvisor.ClusterRole.yaml
```diff
@@ -0,0 +1,15 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  labels:
+    app: cadvisor
+    category: rbac
+    deploy: sourcegraph
+    sourcegraph-resource-requires: no-cluster-admin
```
uwedeportivo, line 8: this requires cluster-admin

base/cadvisor/cadvisor.ClusterRoleBinding.yaml
```diff
@@ -0,0 +1,18 @@
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  labels:
+    app: cadvisor
+    category: rbac
+    deploy: sourcegraph
+    sourcegraph-resource-requires: no-cluster-admin
```
uwedeportivo, line 8: this requires cluster-admin

base/cadvisor/cadvisor.ClusterRole.yaml
uwedeportivo:
```suggestion
    sourcegraph-resource-requires: cluster-admin
```

base/cadvisor/cadvisor.ClusterRoleBinding.yaml
uwedeportivo:
```suggestion
    sourcegraph-resource-requires: cluster-admin
```

base/cadvisor/cadvisor.PodSecurityPolicy.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  labels:
+    app: cadvisor
+    deploy: sourcegraph
+    sourcegraph-resource-requires: no-cluster-admin
```
uwedeportivo, line 7: i think this one too but i'm not sure

overlays/bases/deployments/kustomization.yaml
```diff
@@ -52,3 +52,7 @@ resources:
   - base/precise-code-intel/bundle-manager.Service.yaml
   - base/precise-code-intel/worker.Deployment.yaml
   - base/precise-code-intel/worker.Service.yaml
+  - base/cadvisor/cadvisor.DaemonSet.yaml
```
uwedeportivo, line 55: i think we need to move all of them to rbac-roles, so non-privileged can avoid this whole cadvisor

base/cadvisor/cadvisor.ClusterRole.yaml
bobheadxi: oh hm I saw that prometheus had some conditions in the no-privilege overlay so I used the same labels and whatnot, thanks!

overlays/bases/deployments/kustomization.yaml
bobheadxi: looks like this means we don't have to do any trimming out of cadvisor? 15fa9cc

base/cadvisor/cadvisor.DaemonSet.yaml
bobheadxi: ecaff26  and https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/111 :)

bobheadxi:
after much fenangling with big yaml... emotional moment

image

PR #714, opened 2020-05-28
Update .gitignore for jetbrains IDEs

daxmc99:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ generated/
 Pulumi.*.yaml
 /pkg/gen/openapi-specs
 generated-cluster/
+.idea/


slimsag:
Trying to find a good link for this, but I can't seem to quickly. There was some broader discussion about this in the Go issue tracker somewhere. TL;DR:

I think most editor files should not be committed (e.g. the entire .idea folder), and this includes .gitignore entries for such editor files. I believe this because.. where do we draw the line? should we also add .gitignore entries for sublime, vs code, and other editors or tools people might run in each of our repos? I think then we start ending up with a gitignore template and needing tooling to keep it in sync across all repos :)

A better approach I've seen others use is the global `~/.gitignore` file or `.git/info/exclude` (which is the same as .gitignore but not committed to the repo)

daxmc99:
That makes more sense. TIL

PR #722, opened 2020-06-05
cadvisor: apply args to reduce output

bobheadxi:
Sister [deploy-sourcegraph-docker](https://github.com/sourcegraph/deploy-sourcegraph-docker) change: https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/118


base/cadvisor/cadvisor.DaemonSet.yaml
```diff
@@ -27,7 +27,11 @@ spec:
       - name: cadvisor
         image: index.docker.io/sourcegraph/cadvisor:insiders@sha256:d25e28d7af3be79e4c946682c7fdebf8cde1934212e0ab1150b16cea7514a1e9
         args:
-          - -port=48080
+        # disable container labels to allow whitelisting to reduce noise
```
slimsag, line 30:
```suggestion
        # Kubernetes-specific flags below (other flags are baked into the Docker image)
        #
        # disable container labels to allow whitelisting to reduce noise
```

PR #783, opened 2020-07-03
update advice for --prune flag

davejrt:
Add note to advise users that --prune is destructive and should always be applied to base directory. Address concern in sourcegraph/sourcegraph#11913

### DIFF
--- a/kubectl-apply-all.sh
+++ b/kubectl-apply-all.sh
@@ -6,8 +6,8 @@
 #   * When the cluster is first created
 #   * Whenever the configuration for any resource has been updated
 #
-# This --prune flag is destructive and should always be used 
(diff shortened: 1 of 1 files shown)


kubectl-apply-all.sh
```diff
@@ -5,6 +5,10 @@
 # This file should be run:
 #   * When the cluster is first created
 #   * Whenever the configuration for any resource has been updated
-
+#
+# This --prune flag is destructive and should always be used
```
uwedeportivo, line 9:
```suggestion
# The --prune flag is destructive and should always be used 
```

uwedeportivo: also the label is crucial. it doesn't work well without a label. that's how it computes the diffs of what to keep and what to prune.

PR #784, opened 2020-07-07
prometheus: add builtin alertmanager, labels.level for builtin alerts

bobheadxi:
Add builtin Alertmanager as a target, and add a label for each builtin alert:

* `level: critical` to replace `severity: page`
* `level: warning` otherwise

I believe `severity: page` is used for our current dot-com opsgenie pages, will probably not apply the `severity: page` removal immediately to dot-com (https://github.com/sourcegraph/deploy-sourcegraph-dot-com/pull/2839)

For https://github.com/sourcegraph/sourcegraph/issues/5370

Sister [deploy-sourcegraph-docker](https://github.com/sourcegraph/deploy-sourcegraph-docker) change: n/a

### DIFF
--- a/base/prometheus/prometheus.ConfigMap.yaml
+++ b/base/prometheus/prometheus.ConfigMap.yaml
@@ -244,6 +244,7 @@ data:
             for: 10m
             labels:
               level: critical
+              severity: page
             annotations:
               description: 'Pods missing from {{`{{`}} $labels.app {{`}}`}}: {{`{{`}} $value
             {{`}}`}}'
@@ -254,6 +255,7 @@ data:
             for: 2m
             labels:
               level: critical
+              severity: page
             annotations:
               description: 'No pods are running for {{`{{`}} $labels.app {{`}}`}}: {{`{{`}}
             $value {{`}}`}}'
@@ -264,6 +266,7 @@ data:
               > 20
             labels:
               level: critical
+              severity: page
             annotations:
               description: 'Page load latency > 20s (90th percentile over all routes; current
             value: {{`{{`}}$value{{`}}`}}s)'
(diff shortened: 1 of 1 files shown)


bobheadxi:
dot-com mirrors some of these changes: https://github.com/sourcegraph/deploy-sourcegraph-dot-com/pull/2839

base/prometheus/prometheus.ConfigMap.yaml
```diff
@@ -299,19 +312,23 @@ data:
           - alert: GitserverDiskSpaceLowCritical
             expr: src_gitserver_disk_space_available / src_gitserver_disk_space_total < 0.05
             labels:
-              severity: page
+              level: critical
```
slimsag, line 315: Can you clarify what this does for me? I thought that `severity: page` made these page OpsGenie via our current deployment, so I think this change would mean that these alerts (which we currently rely on to keep sourcegraph.com healthy) would stop paging us via OpsGenie

bobheadxi: I haven't applied this change in https://github.com/sourcegraph/deploy-sourcegraph-dot-com/pull/2839 yet with this in mind, but my thinking is long-term, these alerts will stay and be treated as just another `critical` alert, with all `critical` alerts going to our OpsGenie

slimsag: There is an expectation that the contents of this repository are always the authoritative source and the dot-com repository is merely additions / edits to that; in the spirit of that I would say that changing this should wait until it can be done in the dot-com repository as well

(I like the idea and think it is good otherwise)

bobheadxi: makes sense! f5398fdf adds the `severity: page` back

PR #790, opened 2020-07-15
repo-updater: bump cpu, memory

bobheadxi:
corresponds to https://github.com/sourcegraph/deploy-sourcegraph-dot-com/pull/2975

Sister [deploy-sourcegraph-docker](https://github.com/sourcegraph/deploy-sourcegraph-docker) change: n/a, there are already similar resources configured there https://github.com/sourcegraph/deploy-sourcegraph-docker/blob/master/deploy-repo-updater.sh

### DIFF
--- a/base/repo-updater/repo-updater.Deployment.yaml
+++ b/base/repo-updater/repo-updater.Deployment.yaml
@@ -38,10 +38,10 @@ spec:
           name: debug
         resources:
           limits:
-            cpu: 100m
-            memory: 500Mi
+            cpu: "4"
+            memory: 4Gi
           requests:
-            cpu: 100m
+            cpu: "1"
             memory: 500Mi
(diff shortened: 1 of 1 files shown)


bobheadxi:
> I think and 8x limit is too much and we should revisit this in the future.

I agree - now hovering at <10% usage in dot-com:

image

cc @slimsag

keegancsmith:
Looking at this PR I'd expect the suggestion of `1 CPU, 2Gi` is fine. For truly large instances they can scale it up.

Repeating conversation from https://github.com/sourcegraph/deploy-sourcegraph-dot-com/pull/2975

>> Note: repo-updater does much less on sourcegraph.com so doesn't need as much memory as customers. 4gb is likely overkill (at the moment).

> what does it do differently

everything essentially. repo-updater is memory hungry for customers. It keeps several in memory data structures related to all repos on sourcegraph. Additionally it will periodically source all repos from remote code hosts into memory and do an in memory diff to what sourcegraph has.

> should I adjust the numbers?

Rather than using sourcegraph.com for guidance on resource use for customers, its likely better to use our dogfood cluster / large customer for guidance.

PR #802, opened 2020-07-23
search: generous timeout for search pods

keegancsmith:
These services when searching will use a lot of CPU and IO. This may
make them slow to respond. The readiness probe failing has disasterous
consequences, since k8s' service discovery will remove the pod from the
endpoint list leading to us rebalancing.

This is a guess at a root cause of problems we are seeing on a very
large instance.

### DIFF
--- a/base/indexed-search/indexed-search.StatefulSet.yaml
+++ b/base/indexed-search/indexed-search.StatefulSet.yaml
@@ -35,6 +35,7 @@ spec:
             port: http
             scheme: HTTP
           periodSeconds: 1
+          timeoutSeconds: 5
         resources:
           limits:
             cpu: "2"
--- a/base/searcher/searcher.Deployment.yaml
+++ b/base/searcher/searcher.Deployment.yaml
@@ -50,6 +50,7 @@ spec:
             port: http
             scheme: HTTP
           periodSeconds: 1
+          timeoutSeconds: 5
         resources:
           limits:
             cpu: "2"


keegancsmith:
WDYT about increasing failureThreshold? I don't have great intuitions around this. My guess is we only ever timeout, not fail the health check once the instance is ready.

uwedeportivo:
your hunch here is spot on imho. we shouldn't be so quick in declaring a member of the stateful set as failed because of the repercussions you described. so yeah, i agree, we probably should increase failureThreshold

pecigonzalo:
@keegancsmith Is it correct to assume then that HTTP and `/heatlthz` should start working before it finishes loading/indexing?

keegancsmith:
> @keegancsmith Is it correct to assume then that HTTP and `/heatlthz` should start working before it finishes loading/indexing?

Right now we only start up the HTTP server once loading/indexing is done. Some future work to correctly use readinessProbes/etc so that we can start the HTTP server straight away (such that we can get metrics exported / debug endpoints).

> your hunch here is spot on imho. we shouldn't be so quick in declaring a member of the stateful set as failed because of the repercussions you described. so yeah, i agree, we probably should increase failureThreshold

On further thought I would be very surprised at a non-timeout based failure, so I'll actually just stick with increasing timeout and not failure threshold.

I just checked prod, and I see we sometimes have readiness checks failing, only ever due to timeout. We likely have this problem in some of our other services as well.

PR #806, opened 2020-07-28
Drop unneeded deployments from deploy-sourcegraph

daxmc99:


slimsag:
I'm fine with removing configure/lsif-server-migrator assuming @sourcegraph/code-intel is

configure/nginx-svc is used by some customers and should not be removed.

daxmc99:
Updated this PR to only drop the lsif-migration deployment

slimsag:
@daxmc99 can we merge this?

PR #809, opened 2020-07-31
add app label for prometheus, alertmanager

bobheadxi:
closes https://github.com/sourcegraph/sourcegraph/issues/12482

Sister [deploy-sourcegraph-docker](https://github.com/sourcegraph/deploy-sourcegraph-docker) change: n/a

### DIFF
--- a/base/prometheus/prometheus.ConfigMap.yaml
+++ b/base/prometheus/prometheus.ConfigMap.yaml
@@ -234,14 +234,18 @@ data:
         action: replace
         target_label: kubernetes_pod_name
 
-    - job_name: 'builtin-alertmanager'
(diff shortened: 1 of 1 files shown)


daxmc99:
What are we using app labels for in Prometheus?

bobheadxi:
@daxmc99 there's more context in https://github.com/sourcegraph/sourcegraph/issues/12482, but it seems we used this in the past for the "no pods available" alerts - it doesn't seem like this is available for prometheus

Another possibility is to just not have these alerts for prometheus, because if prometheus is down there won't be alerts anyway but... feels more complete?

daxmc99:
Yup, looks good

PR #815, opened 2020-08-07
overlays: collect PVCs into a separate base

uwedeportivo:
collects PVC resources into a separate base. this allows kustomizations to pick and choose which PVCs they want or avoid creating them altogether. we have customers that need this because they explicitly create PVs and map them to deployments/statefulsets with named claim refs.

(kustomize does not allow deletes)


pecigonzalo:
Wouldn't this be a problem anyway because we specify the PVC in the volume definition of the Deployment?

uwedeportivo:
it assumes the overlay the customer uses also replaces/edits the mention in the deployment

PR #845, opened 2020-09-02
update renovate to use insiders for sourcegraph docker images

bobheadxi:
This (hopefully) configures renovate to set up a PR with insiders images for this repo as part of https://github.com/sourcegraph/sourcegraph/issues/13122 . Automerge is currently not enabled so that I can verify this.

I'm hoping, along with https://github.com/sourcegraph/sourcegraph/pull/13449, that we can get to a state where commits to Sourcegraph land in this repo and then to downstream repos like https://github.com/sourcegraph/deploy-sourcegraph-dogfood-k8s-2 relatively quickly.

**Generating releases**: I'm not entirely sure how we generate releases yet with this new setup - I think we'll have to branch and update images and tag there

Sister [deploy-sourcegraph-docker](https://github.com/sourcegraph/deploy-sourcegraph-docker) change: This change was not planned for `deploy-sourcegraph-docker` as part of https://github.com/sourcegraph/sourcegraph/issues/13122 , but I think it would be possible to adapt it over there if this works out

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -7,11 +7,18 @@
   "prHourlyLimit": 0,
   "packageRules": [
     {
-      "groupName": "Sourcegraph Docker images",
+      "groupName": "Sourcegraph Docker insiders images",
       "packagePatterns": ["^index.docker.io/sourcegraph/"],
-      "ignoreUnstable": false,
       "followTag": "insiders",
-      "extends": ["default:automergeAll"]
+      "ignoreUnstable": false,
+      "automerge": false
+    },
+    {
+      "groupName": "Sourcegraph Docker release images",
+      "packagePatterns": ["^index.docker.io/sourcegraph/"],
+      "versionScheme": "semver",
+      "ignoreUnstable": false,
+      "automerge": false
     },
     {
       "groupName": "Pulumi NPM packages",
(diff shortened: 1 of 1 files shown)


pecigonzalo (approved):
LGTM, but I think there are some concerns as we ask customers to fork and rebase from this source that they will end up pointing to insiders.

bobheadxi:
> as we ask customers to fork and rebase from this source that they will end up pointing to insiders.

I can't find this now, but I swear I remember one of the steps being "check out the appropriate release" - I believe if they do so things should be fine still?

bobheadxi:
To help keep release workflows similar to before, I've added back release renvoate in https://github.com/sourcegraph/deploy-sourcegraph/pull/845/commits/fcfa0c69b0e93b1486844913f73114bb951feb0d - so the workflow will be the same as before for past releases (find the PR to update to release images -> tag that)

uwedeportivo:
we tell customers to use a branch for the appropriate release, not HEAD of main. so i think this is no problem.

uwedeportivo (approved):

bobheadxi:
Merging - both of the update groups are set to _not_ automerge, so we can still keep an eye on this

PR #856, opened 2020-09-08
fix duplicate component label in jaeger component

uwedeportivo:
label was declared twice and YAML decoders complain (but not k8s weirdly enough)


base/jaeger/jaeger.Deployment.yaml
```diff
@@ -8,7 +8,6 @@ metadata:
     app.kubernetes.io/component: jaeger
     app: jaeger
     app.kubernetes.io/name: jaeger
-    app.kubernetes.io/component: all-in-one
```
ggilmore, line 11: @uwedeportivo Are you sure that jaeger isn't relying on this?

ggilmore (approved):

base/jaeger/jaeger.Deployment.yaml
uwedeportivo: i'm not sure. i googled a little bit and didn't find anything. the jaeger k8s operator might use it. not sure

ggilmore: Okay. We should test to make sure that this change is safe .

uwedeportivo: i tested with a test cluster and it seems fine. jaeger pods came up, i could hit the UI

PR #1195, opened 2020-11-02
indexed-search: webserver with a higher watchdog threshold

keegancsmith:
This commit is based off v3.21.2, but changes the zoekt version to have watchdog instrumentation and tuneables.

It tunes the watchdog to trigger after 10 failed attempts, instead of the default 3.

When upgrading to 3.22 (when released), you can switch back to the official images. If you want to keep the watchdog tuning, leave the environment variables in. They will be valid in 3.22.

We encourage you to first upgrade to 3.21, then to apply this change.

You can checkout this branch with `git fetch --all && git checkout k/3.21-watchdog`

### DIFF
--- a/base/indexed-search/indexed-search.StatefulSet.yaml
+++ b/base/indexed-search/indexed-search.StatefulSet.yaml
@@ -23,7 +23,16 @@ spec:
     spec:
       containers:
       - env:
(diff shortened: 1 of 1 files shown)


keegancsmith:
Ignore this PR, contains wrong insiders image.

PR #1316, opened 2020-11-18
Add minio to overlays

pecigonzalo:
This should fix the tests in this repo, and actually correct the generate overlay scripts as minio was not added and it was not being generated/deployed.

We should review and fix this tests/overlay as its quite messy at the moment and tricky to do correctly or even debug.


bobheadxi (approved):
curious how we missed this when minio got first merged :(

bobheadxi:
> curious how we missed this when minio got first merged :(

@pecigonzalo clarified that this is because the code-intel worker did not have a hard dependency on minio, and a hard dependency might have been introduced the failure to deploy minio revealed itself

PR #1323, opened 2020-11-18
Add 1.18 k8s

daxmc99:
Add 1.18 cluster


bobheadxi (approved):
seems fine but does this increase CI time? should we drop one of the older versions?

PR #1332, opened 2020-11-19
Run tests in parallel

daxmc99:
Test improvement


daxmc99:
The Pulumi Go framework also calls t.Parallel()

PR #1381, opened 2020-11-28
feat(scripts): add a script to clean-up all kubernetes resources

mtmn:
Summary:
There is `kubectl-apply-all.sh` already in place, however, there should
be a "delete-all" script present for more smooth playground scenarios in
real clusters.

### DIFF
--- a/kubectl-delete-all.sh
+++ b/kubectl-delete-all.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+#
+# This file should contain `kubectl delete` commands for all of your configured resources.
+#
+# This file should be run:
+#   * When you want to clean-up all resources created
+#
+# Remove the base Soucegraph deployment
+# shellcheck disable=SC2068
+kubectl delete -f base --recursive $@


uwedeportivo:
i would rather not offer a script like this or at least have a "are you sure prompt". we should probably also change the name. kubectl-apply and kubectl-delete are close ('a' and 'd' are close on the keyboard and with shell completion...).
an alternative would be to put the script contents in the admin docs in a delete cluster section. then it's a clearly intentional action on the customer's part.

mtmn:
@uwedeportivo - now that I think of it, it makes much more sense to update documentation rather than add a potentially malicious script in here.

is it possible to contribute to documentation as well?

daxmc99:
I think deploying Sourcegraph to a namespace and then simply deleting that namespace should suffice for most playground needs.
Our documentation could be improved around this though.

uwedeportivo:
@mtmn yes, of course. and thanks for contributing. it would be somewhere in https://github.com/sourcegraph/sourcegraph/tree/main/doc/admin

mtmn:
@uwedeportivo - thank you, will create a documentation entry.

PR #1382, opened 2020-11-29
tools: generate update-docker-tags args from sourcegraph/sourcegraph

bobheadxi:
Generates update-docker-tag args from the master list over in `sourcegraph/sourcegraph` (https://github.com/sourcegraph/sourcegraph/pull/16227) instead of requiring manual maintenance. Closes https://github.com/sourcegraph/sourcegraph/issues/15896

As a reminder, we can't just wildcard `sourcegraph/*` images during version enforcement because we have images that don't conform to our standard versioning (e.g. `sourcegraph/alpine`)

The entrypoint remains the same, so this will not break our tooling/CI that now depends on the script.

### Checklist

* [x] Sister [deploy-sourcegraph-docker](https://github.com/sourcegraph/deploy-sourcegraph-docker) change: https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/187

### DIFF
--- a/tools/enforce-tags/main.go
+++ b/tools/enforce-tags/main.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/sourcegraph/sourcegraph/enterprise/dev/ci/images"
+)
+
+func main() {
+	if len(os.Args) != 3 {
+		log.Fatal("constraint, directory arguments must be provided")
+	}
+	var (
+		constraint = os.Args[1]
+		dir        = os.Args[2]
+	)
+	args := []string{
+		"run",
+		"github.com/slimsag/update-docker-tags",
+	}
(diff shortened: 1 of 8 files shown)


tools/enforce-tags/main.go
```diff
@@ -0,0 +1,38 @@
...
+	if len(os.Args) != 3 {
+		log.Fatal("constraint, directory arguments must be provided")
+	}
+	var (
+		constraint = os.Args[1]
+		dir        = os.Args[2]
+	)
+	args := []string{
+		"run",
+		"github.com/slimsag/update-docker-tags",
```
uwedeportivo, line 23: we should probably make this a library and call it directly

uwedeportivo (approved):

tools/enforce-tags/main.go
bobheadxi: that would be nice in the long run - i think a first step would be a better path towards owning this tool ourselves, however: https://github.com/sourcegraph/sourcegraph/issues/16280

PR #1408, opened 2020-12-02
move cadvisor daemonset to deployments

davejrt:


overlays/bases/rbac-roles/kustomization.yaml
```diff
@@ -7,6 +7,5 @@ resources:
   - base/prometheus/prometheus.ClusterRole.yaml
   - base/cadvisor/cadvisor.ClusterRoleBinding.yaml
   - base/cadvisor/cadvisor.ClusterRole.yaml
-  - base/cadvisor/cadvisor.DaemonSet.yaml
```
bobheadxi, line 10: IIRC i put this is here because I'm not sure cadvisor really works without the cluster role thingo (could be wrong though? this was a bit of a shot in the dark) do you mind explaining this move a bit?

davejrt: This isn't removing the cluster role, but if someone wants to make a change to something inside cadvisor (the image for example) they shouldn't need to include `..bases/rbac` in their kustomization, only deployments.

bobheadxi: oh I think I might have misunderstood this back then a bit then

bobheadxi (approved):
lgtm!

overlays/bases/rbac-roles/kustomization.yaml
davejrt: TIL this after trying to find out what my kustomizations weren't generating as expected :-)

pecigonzalo:
Why do we need to do this?

pecigonzalo (approved):

daxmc99:
> ~~Why do we need to do this?~~ My bad, from the title I thought we were moving it from being a DaemonSet to Deployment type.

Ahh, this should have been nested in this other deployment location. I had the same question. Does Kustomize expect deployments there though?

davejrt:
@daxmc99 no, deployments is just the directory name. Hence before this worked when the deployment was actually specified in rbac roles, the same reason we have statefulsets under deployments.

PR #1578, opened 2021-01-07
[superseded] prometheus: add config for filtering cadvisor metrics by namespace

bobheadxi:
This is a follow-up for changes that fix https://github.com/sourcegraph/sourcegraph/issues/17069. Part of the changes is relaxed regex matching for cAdvisor targets, to ensure we have compatibility with docker-compose environments (https://github.com/sourcegraph/sourcegraph/issues/17072). Unfortunately this means in environments like GCP (Cloud and k8s.sgdev), certain containers like GCP's `prometheus-to-*` exporters get picked up too.

I couldn't really find an ergonomic way to do this filtering as part of the monitoring library API and/or regex short of including a hardcoded list of unwanted containers (this might be bad, due to cAdvisor picking up anything and everything, and the nature of the loose regex). Because of this, I think it is probably best to do this _before_ the metrics get ingested in Prometheus in the first place, which is probably what we want to do anyway to avoid excess Prometheus usage (there was a customer-related issue a while back in this vein, https://github.com/sourcegraph/customer/issues/75, that I think this config will be able to fix)

**Note**: This is currently an optional, commented out config. I would like to make it a default option but it doesn't seem like we have consistent namespacing by default everywhere (e.g. Prometheus lives in `default` by default) - this PR also changes that, but might need to revert if there's some context I'm missing here. Also, because namespace changes are recommended to be applied via kustomize, it might be possible that customers might forget to change the regex value added in this PR. Open to thoughts though

Alternative approaches attempted include filtering on the monitoring generator side (not very nice, as mentioned above), and using relabel config that tries to ensure `container_label_io_kubernetes_pod_namespace` and `ns` are the same (only get metrics within the same namespace as cAdvisor). I'm not sure it's actually possible - I attempted something like [this unanswered SO question](https://stackoverflow.com/questions/62193593/how-to-keep-matched-source-labels-in-prometheus-relabel-config) that did not work.

### Checklist


bobheadxi:
This change is already live in k8s.sgdev.org and will be applied in Cloud by https://github.com/sourcegraph/deploy-sourcegraph-dot-com/pull/4129

base/cadvisor/cadvisor.ClusterRoleBinding.yaml
```diff
@@ -16,4 +16,4 @@ subjects:
 - kind: ServiceAccount
   name: cadvisor
   # value of namespace needs to match namespace value in base/cadvisor/cadvisor.ServiceAccount.yaml
-  namespace: default
+  namespace: ns-sourcegraph
```
pecigonzalo, line 19: I would make this only `sourcegraph` its implicitly an NS.

pecigonzalo:
Related: https://github.com/sourcegraph/sourcegraph/issues/12950

base/cadvisor/cadvisor.ClusterRoleBinding.yaml
bobheadxi: `ns-sourcegraph` is the default: https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/sourcegraph/deploy-sourcegraph%24+namespace:+ns-sourcegraph&patternType=structural

pecigonzalo: That is unfortunate.

daxmc99:
As part of this, we also need to validate that containerd vs docker has the same level of monitoring.

bobheadxi:
> As part of this, we also need to validate that containerd vs docker has the same level of monitoring.

I think this is a different task, as described in the PR body this only pertains to issues with ease of querying of the cAdvisor `name` label.

base/prometheus/prometheus.ConfigMap.yaml
```diff
@@ -238,6 +238,16 @@ data:
...
+      # cAdvisor-specific customization. Drop container metrics exported by cAdvisor
+      # not in the same namespace as Sourcegraph.
+      # Uncomment this if you have problems with certain dashboards or cAdvisor itself
+      # picking up non-Sourcegraph services. Ensure all Sourcegraph services are running
+      # within the Sourcegraph namespace you have defined.
+      # The regex must keep matches on '^$' (empty string) to ensure other metrics do not
+      # get dropped.
+      # - source_labels: [container_label_io_kubernetes_pod_namespace]
+      #   regex: ^$|ns-sourcegraph # ensure this matches with namespace declarations
+      #   action: keep
```
bobheadxi, line 250: I kind of want to keep this enabled as a default, but the downside is it might get missed when applying the namespace kustomize

pecigonzalo (approved):

base/prometheus/prometheus.ClusterRoleBinding.yaml
```diff
@@ -15,4 +15,4 @@ subjects:
 - kind: ServiceAccount
   name: prometheus
   # value of namespace needs to match namespace value in base/prometheus/prometheus.ServiceAccount.yaml
-  namespace: default
+  namespace: ns-sourcegraph
```
davejrt, line 18: Can you explain why this gets changed from the `default` namespace to `ns-sourcegraph` when the rest of sourcegraph is still set to deploy into the `default` namespace? I know **if** a customer uses kustomize it gets changed, but wouldn't it be better to change this namespace with kustomize as well, instead of having the namespaces diverge like this?

bobheadxi: this namespace is the default namespace used in the repo currently: https://github.com/sourcegraph/deploy-sourcegraph/pull/1578#discussion_r553379376

davejrt: Only for some overlays...not all of them. For example non-root overlays don't apply a different ns. If you clone this repo, and apply the base as is, everything else including the prometheus deploymet is deployed into default. 

My suggestion would be something like this:
https://github.com/sourcegraph/deploy-sourcegraph/commit/6d2b115eb7e5402662e8e0bfa7105ca8ec8c28a1

bobheadxi: Ah I see! Completely misunderstood this, thanks!

Looking at your patch, does this mean we have to ensure both the base and overlay are kept in-sync manually?

davejrt: I looked into kustomize and I can't see a way to edit one line in otherwise equal configmap, so unfortunately, yes we'd need to keep this in sync "manually" but the only change it the namespace on that one line. I don't see why we couldn't just do a find/replace on that line in some automated fashion

bobheadxi: I think that's a sufficient solution as we hold out for Dhall - do you mind pushing your patch to this branch?

also cc @pecigonzalo for a look again

bobheadxi: @davejrt I cherry-picked your change here: [`93e5188` (#1578)](https://github.com/sourcegraph/deploy-sourcegraph/pull/1578/commits/93e5188d660bfd66bf8de56cc7004c5b07e7c104)

pecigonzalo: Just looking at this in its uncommented format, wouldn't this break monitoring for anyone deploying into an implicit namespace (`kubectl ns foo && kubectl apply` )?

davejrt: I think it's better off actually deleting this altogether, as we don't explicitly specify a namespace anywhere else in our manifests. If someone uses kustomize to change the namespace, these clusterrole bindings will be changed. If they apply to an alternate namespace implicitly as per @pecigonzalo example, they aren't affected.

bobheadxi: > I think it's better off actually deleting this altogether

this configuration is pretty important to have in most environments, and this PR revolves around adding this bit - see description.

I'm curious as to why we don't deploy to a namespace by default? Seems ideal to be able to identify Sourcegraph services separately from whatever else is running in a customer's environment by default

bobheadxi: I did some digging, and the following Kubernetes metadata is the only thing we can get on a per-container label level (which seems to be the only level at which is exported by cadvisor):

- `io.kubernetes.container.name`
- `io.kubernetes.pod.name`
- `io.kubernetes.pod.namespace`
- `io.kubernetes.pod.uid`

`name` alone is not sufficient, as described

bobheadxi: an alternative solution is that cAdvisor can be configured to export certain container env variables as labels. We can then require that all containers that want to be monitored should set the following env:

```yml
        env:
        - name: CADVISOR_SHOULD_MONITOR
          value: true
```

bobheadxi: removed the namespace declarations in 1aa44df8f2100967885eea183bc77f5a91821f49

pecigonzalo: > I'm curious as to why we don't deploy to a namespace by default? Seems ideal to be able to identify Sourcegraph services separately from whatever else is running in a customer's environment by default

We do want customers to deploy in a namespace, but the name and such is generally defined by the end-user in most cases as they might have some naming conventions like `ns-<name>` or `prod-<name` or a combination of those or others.
If we add the namespace to all resources, the customer would have to either create the same name on its own, or we will have to bundle a Namespace resoruces. In both cases, they will have to modify all files in many places to make them match if they dont use the bundled one.

Kustomize/Dhall resolve this last part AFAIK but not the templating in the config file. Maybe the pod can template the file on startup with an init container or process using the `KUBE_NAMESPACE` env var.

bobheadxi: Do we want to explore a long-term solution to this? I think I've already sunk a huge amount of time trawling forums and docs and even source code for both the original cAdvisor issues and ergonomic ways to try and resolve this filtering problem

I think there are a lot of solutions here including something like:

> Maybe the pod can template the file on startup with an init container or process using the KUBE_NAMESPACE env var.

I have some other ideas as well, but they all seem pretty hack and could probably do with a more detailed comparison of how each method will work out _and_ how they might map to dhall. I've also been pondering whether we want to stick with cAdvisor at all, maybe there are more modern collectors out there with the versatility we need?

I'm tempted right now to just revert this PR to the first commit (f886f2fab31c3b313833e9c94db956061f59a55c) and merge that, and let customers know how to add this config if they encounter issues

pecigonzalo: I think that sounds good and lets just open or keep the original issue for this to track those other ideas, so we dont start from scratch.

bobheadxi: Opened https://github.com/sourcegraph/deploy-sourcegraph/pull/1644 , will open an issue for this stuff

bobheadxi: Follow-up: https://github.com/sourcegraph/sourcegraph/issues/17365

bobheadxi:
An issue for exploring alternatives has been opened here: https://github.com/sourcegraph/sourcegraph/issues/17365

Closing this PR!

PR #1586, opened 2021-01-08
Update ingress 

daxmc99:
Fixes https://github.com/sourcegraph/sourcegraph/issues/16808

- upgrade to 0.43.0 of ingress
- Add code-intel db to minikube kustomization
- Upgrade Ingress resource to v1
- Don't actually upgrade to v1beta1 since that is only available in k8s 1.19+ per https://github.com/kubernetes/kubernetes/issues/90077#issuecomment-718778960 

>networking.k8s.io/v1beta1 == 1.14 to 1.18
>
>networking.k8s.io/v1 = 1.19+

### Checklist

* [x] Sister [deploy-sourcegraph-docker](https://github.com/sourcegraph/deploy-sourcegraph-docker) : change: k8s only change 

### DIFF
--- a/configure/ingress-nginx/mandatory.yaml
+++ b/configure/ingress-nginx/mandatory.yaml
@@ -17,7 +17,6 @@ metadata:
     app.kubernetes.io/name: ingress-nginx
     app.kubernetes.io/instance: ingress-nginx
     app.kubernetes.io/version: 0.43.0
-    app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/component: controller
   name: ingress-nginx
   namespace: ingress-nginx
@@ -31,11 +30,9 @@ metadata:
     app.kubernetes.io/name: ingress-nginx
     app.kubernetes.io/instance: ingress-nginx
     app.kubernetes.io/version: 0.43.0
-    app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/component: controller
   name: ingress-nginx-controller
   namespace: ingress-nginx
-data:
 ---
 # Source: ingress-nginx/templates/clusterrole.yaml
 apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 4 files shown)


configure/ingress-nginx/mandatory.yaml
```diff
@@ -1,63 +1,56 @@
...
 kind: ClusterRole
 metadata:
-  name: nginx-ingress-clusterrole
   labels:
+    helm.sh/chart: ingress-nginx-3.19.0
     app.kubernetes.io/name: ingress-nginx
-    app.kubernetes.io/part-of: ingress-nginx
+    app.kubernetes.io/instance: ingress-nginx
+    app.kubernetes.io/version: 0.43.0
+    app.kubernetes.io/managed-by: Helm
```
pecigonzalo, line 49:
```suggestion
```

I understand this comes from Helm, but its not managed by it in our deployment.

configure/ingress-nginx/mandatory.yaml
```diff
@@ -1,63 +1,56 @@
+
 apiVersion: v1
 kind: Namespace
 metadata:
   name: ingress-nginx
   labels:
     app.kubernetes.io/name: ingress-nginx
-    app.kubernetes.io/part-of: ingress-nginx
+    app.kubernetes.io/instance: ingress-nginx
```
pecigonzalo, line 8: I believe this remains `part-of` according to https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/.

configure/ingress-nginx/mandatory.yaml
pecigonzalo: I will not comment one very instance to keep the discussion here. I would remove all `managed-by` and `helm.sh/chart` references.

pecigonzalo (approved):
LGTM in general. My only concern is how we are going to cleanup all resources that are not part of the updated one.

daxmc99:
For the upgrades, I'm not entirely sure what we should do there. A helm based install might use pre-install hooks. 
I think updating docs to mention that we upgraded the Nginx ingress controller should be enough. A user could delete the namespace to cleanup the resources. 
In our internal environments, I expect this to apply without any issues.

PR #1612, opened 2021-01-12
Switch back to wrouesnel postgres exporter

daxmc99:
The codeintel DB cannot use the sourcegraph/postgres-exporter image today since it does not contain a migration table which our the sourcegraph/postgres-exporter queries for. 

This causes the exporter for code intel to continually emit errors

Found during https://github.com/sourcegraph/sourcegraph/pull/17131

### Checklist

* [x] Sister [deploy-sourcegraph-docker](https://github.com/sourcegraph/deploy-sourcegraph-docker) change: We don't deploy Postgres exporter in docker-compose

### DIFF
--- a/base/codeintel-db/codeintel-db.Deployment.yaml
+++ b/base/codeintel-db/codeintel-db.Deployment.yaml
@@ -64,6 +64,7 @@ spec:
       - env:
         - name: DATA_SOURCE_NAME
           value: postgres://sg:@localhost:5432/?sslmode=disable
+        # Dax: Temporarily switch back to upstream postgres exporter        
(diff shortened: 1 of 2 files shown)


uwedeportivo (approved):
this is temporary, right ? we do want our image eventually, correct ? we could configure it to not query that missing table, i think.

daxmc99:
Yeah, with a release coming soon I wanted to revert. 
There are two approaches of either removing the config from the image and having it external or we could overwrite the config for the codeintel exporter specifically (including with the default config). 
We originally did this to avoid needing to provide configuration values everywhere we used the Postgres exporter.

bobheadxi:
I think we should include an explanation as a comment here

PR #1851, opened 2021-02-16
add codeinsights-db (TimescaleDB) deployment

slimsag:
This PR adds the Code Insights TimescaleDB deployment, which will be used to
store time series data for tracking insights about code. For more information
on what this will be used for, see https://github.com/sourcegraph/sourcegraph/issues/17218

TimescaleDB is a plugin for Postgres which extends it to support time series
data storage. We only use the OSS version, and so deploying it is always free
(and their non-free version supports our use case anyway.)

People deploying Sourcegraph should use this deployment and not try to deploy
TimescaleDB as a plugin in their Postgres deployment for a few reasons:

1. TimescaleDB only supports specific Postgres versions, e.g. not Postgres 13
   yet.
2. TimescaleDB should have isolated resources from the rest of Sourcegraph, so
   as to prevent Code Insights from interfering with other Sourcegraph features
   in general.

Signed-off-by: Stephen Gutekanst <stephen@sourcegraph.com>

### Checklist

* [ ] Sister [deploy-sourcegraph-docker](https://github.com/sourcegraph/deploy-sourcegraph-docker) change: I will send this soon.

### DIFF
--- a/base/codeinsights-db/codeinsights-db.Deployment.yaml
+++ b/base/codeinsights-db/codeinsights-db.Deployment.yaml
@@ -55,17 +55,18 @@ spec:
       - env:
         - name: DATA_SOURCE_NAME
           value: postgres://sg:@localhost:5432/?sslmode=disable
-        # Dax: Temporarily switch back to upstream postgres exporter        
-        image: wrouesnel/postgres_exporter:v0.7.0@sha256:785c919627c06f540d515aac88b7966f352403f73e931e70dc2cbf783146a98b
-        terminationMessagePolicy: FallbackToLogsOnError
-        name: pgsql-exporter
-        resources:
-          limits:
-            cpu: 10m
-            memory: 50Mi
-          requests:
-            cpu: 10m
-            memory: 50Mi
+        # Dax: Temporarily switch back to upstream postgres exporter
+        # https://github.com/sourcegraph/sourcegraph/issues/18225
+        # image: wrouesnel/postgres_exporter:v0.7.0@sha256:785c919627c06f540d515aac88b7966f352403f73e931e70dc2cbf783146a98b
+        # terminationMessagePolicy: FallbackToLogsOnError
+        # name: pgsql-exporter
+        # resources:
+        #   limits:
+        #     cpu: 10m
+        #     memory: 50Mi
+        #   requests:
+        #     cpu: 10m
+        #     memory: 50Mi
       securityContext:
         runAsUser: 0
       volumes:
--- a/base/codeinsights-db/codeinsights-db.Deployment.yaml
+++ b/base/codeinsights-db/codeinsights-db.Deployment.yaml
@@ -4,6 +4,7 @@ metadata:
   annotations:
     description: Code Insights TimescaleDB instance.
   labels:
+    app.kubernetes.io/component: codeinsights-db
     deploy: sourcegraph
     sourcegraph-resource-requires: no-cluster-admin
   name: codeinsights-db
--- a/base/codeinsights-db/codeinsights-db.Deployment.yaml
+++ b/base/codeinsights-db/codeinsights-db.Deployment.yaml
@@ -1,0 +1,77 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    description: Code Insights TimescaleDB instance.
+  labels:
+    deploy: sourcegraph
+    sourcegraph-resource-requires: no-cluster-admin
+  name: codeinsights-db
+spec:
+  minReadySeconds: 10
+  replicas: 1
+  revisionHistoryLimit: 10
+  selector:
+    matchLabels:
+      app: codeinsights-db
+  strategy:
+    type: Recreate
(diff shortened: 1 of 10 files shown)


slimsag:
Note:

* I based this on codeintel-db, so as to make sure I captured the needed non-root overlays.
* The postgres.conf included here comes from TimescaleDB itself, I included comments in the ConfigMap about differences and how to use `timescaledb-tune` to generate a good config file.

Questions:

* What do I need to do to land this on https://github.com/sourcegraph/deploy-sourcegraph-dot-com/ ? I am out of touch with how we do this today.
* Should I also send a PR to https://github.com/sourcegraph/deploy-sourcegraph-core ?

davejrt (approved):

pecigonzalo:
> What do I need to do to land this on https://github.com/sourcegraph/deploy-sourcegraph-dot-com/ ? I am out of touch with how we do this today.

`dot-com` uses CloudSQL, so it needs to be added as an extension through Terraform but it seems like it's not supported on CloudSQL, RDS, etc. This could be a problem for our Cloud and for certain customers using external services for their DB, unless we require this service to be run as a pod.

https://github.com/timescale/timescaledb/issues/65

> Should I also send a PR to https://github.com/sourcegraph/deploy-sourcegraph-core ?

Not a requirement as that project is parked. If you want to do it low-prio, sure!

uwedeportivo (approved):

slimsag:
@pecigonzalo As I noted in the description, I think people should deploy this as a pod (including us on sourcegraph.com) and not as an extension in their other Postgres deployment or similar. I will see about sending docs to make this very explicit/clear.

We will also only be storing the insights data here, so data loss will be minimal (just insights) should someone fail to backup the pods disks.

slimsag:
Once I merge this, what is the process for getting this into the dot-com repo? Do I perform a `git merge`, does Robert's tooling do that automatically, something else, etc?

ggilmore:
@slimsag Yes, a git merge will be necessary. https://about.sourcegraph.com/handbook/engineering/deployments#merging-upstream-deploy-sourcegraph-into-deploy-sourcegraph-dot-com

Amongst other things that I can think of, you'll need to make sure that the `dot-com` deployment manifest specify the `prod` namespace.

base/codeinsights-db/codeinsights-db.Deployment.yaml
```diff
@@ -0,0 +1,77 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    description: Code Insights TimescaleDB instance.
+  labels:
+    deploy: sourcegraph
+    sourcegraph-resource-requires: no-cluster-admin
```
ggilmore, line 8: Please add `app.kubernetes.io/component` labels to all the manifests here. See https://github.com/sourcegraph/deploy-sourcegraph/blob/0eaab6588cd270cc536beda0101a7ad681d167e6/base/frontend/sourcegraph-frontend.Ingress.yaml#L15

base/codeinsights-db/codeinsights-db.Deployment.yaml
```diff
@@ -0,0 +1,77 @@
...
+        volumeMounts:
+        - mountPath: /data
+          name: disk
+        - mountPath: /conf
+          name: timescaledb-conf
+      - env:
+        - name: DATA_SOURCE_NAME
+          value: postgres://sg:@localhost:5432/?sslmode=disable
+        # Dax: Temporarily switch back to upstream postgres exporter        
+        image: wrouesnel/postgres_exporter:v0.7.0@sha256:785c919627c06f540d515aac88b7966f352403f73e931e70dc2cbf783146a98b
```
ggilmore, line 58: Is this the exporter image that you want to use? If so, please specify a registry in the image url - like `index.docker.io`.

ggilmore (approved):

base/codeinsights-db/codeinsights-db.Deployment.yaml
slimsag: Should I add this to `base/codeintel-db/codeintel-db.Deployment.yaml`? It's not in there either

base/codeinsights-db/codeinsights-db.Deployment.yaml
slimsag: I merely copied this from `base/codeintel-db/codeintel-db.Deployment.yaml` - I will update both though. @daxmc99 is there an issue for whatever the comment means?

daxmc99: The upstream image is needed for talking to databases that do not have the `schema_migrations` table. Our image assumes that the table exists and will fail otherwise.

base/codeinsights-db/codeinsights-db.Deployment.yaml
ggilmore: That would be great if you could, thanks. It's not just just the deployment manifests though, it's all files related to the component. ￼

PR #1852, opened 2021-02-16
add limits for ephemeral storage

davejrt:
These initial figures are speculative at best, and customers would need to set these themselves in my cases I imagine depending on deployment size. 

These change also doesn't modify anything in the configure directory. 

fixes https://github.com/sourcegraph/sourcegraph/issues/9604


uwedeportivo (approved):
LGTM, seems like reasonable values. Ideally anything that requires more would be using an explicit volume mount in our app.

PR #1915, opened 2021-02-25
add missing deployments to minikube overlay

davejrt:


overlays/minikube/kustomization.yaml
```diff
@@ -96,6 +96,24 @@ patchesJson6902:
...
+      name: codeinsights-db
+      group: apps
+      version: v1
+    path: delete-resources.yaml  
+  - target:
+      kind: Deployment
+      name: minio
+      group: apps
+      version: v1
+    path: delete-resources.yaml
```
daxmc99, line 116: I believe this will remove code-intel from the `minikube` overlay? Just wanted to note that

PR #1936, opened 2021-02-26
codeinsights-db: correct data volume path

slimsag:
I failed to verify `/data` was the correct path (it is for our other Postgres images, but not for this TimescaleDB image), which means codeinsights-db data is not being retained currently. I noticed this because the DB was lost on sourcegraph.com (chart looks empty.)

The good news is that I am actively working on building historical data for insights, so this data loss doesn't really matter - it'll all be rebuilt later.

Fixes https://github.com/sourcegraph/sourcegraph/issues/18704

Signed-off-by: Stephen Gutekanst <stephen@sourcegraph.com>

### Checklist

* [x] Sister [deploy-sourcegraph-docker](https://github.com/sourcegraph/deploy-sourcegraph-docker) change: https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/268

### DIFF
--- a/base/codeinsights-db/codeinsights-db.Deployment.yaml
+++ b/base/codeinsights-db/codeinsights-db.Deployment.yaml
@@ -30,6 +30,8 @@ spec:
         env:
         - name: POSTGRES_PASSWORD # Accessible by Sourcegraph applications on the network only, so password auth is not used.
           value: password
+        - name: PGDATA
+          value: /var/lib/postgresql/data/pgdata
         terminationMessagePolicy: FallbackToLogsOnError
         ports:
         - containerPort: 5432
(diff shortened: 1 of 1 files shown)


ggilmore (approved):
@slimsag A thought - why was the deployment still healthy if the path was incorrect?

Also, does this mean that users will start running into ephemeral storage issues soon?

slimsag:
> A thought - why was the deployment still healthy if the path was incorrect?

It was able to write into `/var/lib/postgresql/data` with no problem (ephemeral storage) - I can't see any easy way to pick this up in a health check, but maybe you have an idea?

When the container restarted, it's as if the DB was entirely empty so "there is no issue" either.

> Also, does this mean that existing users will start running into ephemeral storage issues soon?

Luckily this is unlikely, codeinsights is behind a feature flag and of-by-default. Plus, even when enabled the database is rather small unless you add a lot of insights and it records data for a **really** long time (like a year) - people will upgrade before they encounter any ephemeral storage issues I expect.

PR #2073, opened 2021-03-12
Remove status triggers for automerge

pecigonzalo:
This prevents this to constantly trigger as `status` triggers too frequntly.

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -1,4 +1,4 @@
-# based on https://github.com/pascalgn/automerge-action/tree/v0.9.0#usage
+# based on https://github.com/pascalgn/automerge-action/tree/v0.13.0#usage
 name: Automerge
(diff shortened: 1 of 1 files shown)


davejrt:
No objections here, but a question: is the current behaviour that when the automerge label is applied, and the status event occurs, that automerge runs?

davejrt (approved):

davejrt:
Got it, and this just prevents that flow from running every time a status event is triggered on the repo?

pecigonzalo:
Exactly, when doing the renovate/automerge for `deploy-sourcegraph-dot-com` I found some strange behavior of multiple runs for `status` which I could not trace, but likely triggers for all the rebases and there was also some odd endless loop of them running all the time many times per commit.

davejrt:
got it...thanks for clarifying :-)

PR #2117, opened 2021-03-16
update crb, fix dead link

davejrt:
addresses the concerns in https://github.com/sourcegraph/sourcegraph/issues/19077

related docs update https://github.com/sourcegraph/sourcegraph/pull/19171


base/cadvisor/cadvisor.ClusterRoleBinding.yaml
```diff
@@ -15,3 +15,4 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: cadvisor
+  namespace: default
```
pecigonzalo, line 18: Cant we put this in our NS?

pecigonzalo (approved):

PR #2274, opened 2021-04-05
fixes non root deployments

davejrt:
frontend deployments are created without a role/rolebinding, which means the frontend cannot connect to the api correctly. 

In our tests, we explicitly apply with a service account that has been configured in the `test.sh` 

https://sourcegraph.com/github.com/sourcegraph/deploy-sourcegraph/-/blob/tests/integration/restricted/test.sh#L73


daxmc99:
I think this requires additional customization. We need to drop the `clusterRoles` for prometheus & cAdvisor as part of this

PR #2298, opened 2021-04-07
Increase frontend and precise-code-intel-worker replicas

pecigonzalo:
By default we deploy a single replica but these services support
scaling horizontally and can be HA by running 2 replicas.

### Checklist


pecigonzalo:
We could also change https://github.com/sourcegraph/deploy-sourcegraph/pull/2298/files#diff-9c5f43db27e3a718ba6e654777ead81272acea0decbf9209f9411cdca26080c7L21 to set `maxUnavailable` to `0`, but I think its also a good idea to deploy 2 replicas by default.

davejrt (approved):

PR #2344, opened 2021-04-08
use new liveness and readiness endpoint

davejrt:
uses different endpoint for liveness and readiness to avoid crashing during large migrations

related: https://github.com/sourcegraph/sourcegraph/issues/19757

### DIFF
--- a/base/frontend/sourcegraph-frontend.Deployment.yaml
+++ b/base/frontend/sourcegraph-frontend.Deployment.yaml
@@ -75,14 +75,14 @@ spec:
         livenessProbe:
           httpGet:
             path: /healthz
-            port: http
+            port: health-check
             scheme: HTTP
           initialDelaySeconds: 300
           timeoutSeconds: 5
         readinessProbe:
           httpGet:
-            path: /healthz
-            port: http
+            path: /ready
+            port: health-check
             scheme: HTTP
           periodSeconds: 5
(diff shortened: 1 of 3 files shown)


efritz (approved):

base/frontend/sourcegraph-frontend.Deployment.yaml
```diff
@@ -75,14 +75,14 @@ spec:
...
+            port: debug
             scheme: HTTP
           initialDelaySeconds: 300
           timeoutSeconds: 5
         readinessProbe:
           httpGet:
-            path: /healthz
-            port: http
+            path: /ready
+            port: debug
```
keegancsmith, line 85: @daxmc99 @efritz I think the switch from `http` to `debug` means we will start sending traffic to a sourcegraph-frontend before the `http` port is ready. This is often the case, since `debug` starts up straight away, while `http` does some setup/migrations/etc before being ready.

efritz: The ready endpoint should only report true directly before or after the HTTP port is accepting connections.

Pretty garbage endpoint otherwise.

keegancsmith: aah thanks for the clarification! See now that we pass in a channel to the ready endpoint.

PR #2369, opened 2021-04-09
match cache size to request

davejrt:
We should match the size of the cache to the default request for ephemeral storage.


daxmc99 (commented):
I would personally prefer we raise the requests to this value. Decreasing it may have unintended consequences

davejrt:
@daxmc99 I've increased the requests and limits. Do you off hand know if having the request/limit set to the same value would result in this always being evicted? I suspect we should allow some some burst right?

base/symbols/symbols.Deployment.yaml
```diff
@@ -61,11 +61,11 @@ spec:
         resources:
           limits:
             cpu: "2"
-            ephemeral-storage: "8Gi"
+            ephemeral-storage: "12Gi"
             memory: 2G
           requests:
             cpu: 500m
-            ephemeral-storage: "4Gi"
+            ephemeral-storage: "10Gi"
```
daxmc99, line 68:
```suggestion
            ephemeral-storage: "10G"
```

base/symbols/symbols.Deployment.yaml
```diff
@@ -61,11 +61,11 @@ spec:
         resources:
           limits:
             cpu: "2"
-            ephemeral-storage: "8Gi"
+            ephemeral-storage: "12Gi"
```
daxmc99, line 64:
```suggestion
            ephemeral-storage: "12G"
```

daxmc99 (changes requested):
I think this makes the most sense. 
This will make 10GB a requirement to schedule the pod though. 
The pod will only be evicted if it reaches 12 GB 
We should document this in the changelog

PR #2439, opened 2021-04-13
add default sc for minikube

davejrt:
changes the storageclass to the default for minikube, rather than deleting it.


abeatrix (approved):
🤯 👍

PR #2898, opened 2021-04-29
redis: bump memory requests/limits to 7 GB

ggilmore:
cc @mike-r-mclaughlin 

Both redis-cache and redis-store  have `maxmemory 6gb` in their redis.conf: https://github.com/sourcegraph/infrastructure/blob/main/docker-images/redis-cache/redis.conf#L4 and https://github.com/sourcegraph/infrastructure/blob/3d83a92ce64650f46754e2ac55c3accfe84ab0d7/docker-images/redis-store/redis.conf#L4.  According to https://redis.io/topics/lru-cache#eviction-policies, redis will start evicting things from its cache once it hits that limit. However, because our _k8s pod_ memory limits are also set to 6 GB - k8s will kill the redis pods before they have a chance to perform that eviction process. 

Giving redis one more gigabyte to work with should give it enough room to do its cache eviction without being OOMKilled


daxmc99 (approved):
LGTM

PR #2902, opened 2021-05-03
Reduce chance of permission errors

daxmc99:
There have been reports of Postgres failing to start if permissions are not setup correctly.

While we aren't clear on the root cause of these errors, this is a easy fix that customers have reported fix the issue

### Checklist


davejrt (approved):

daxmc99:
This only helps on upgrades from 3.27 to 3.28, should we consider backporting it?

PR #2953, opened 2021-05-11
Set ephemeral storage to match cache

daxmc99:
Related to https://github.com/sourcegraph/deploy-sourcegraph/pull/2369

Fixes https://github.com/sourcegraph/sourcegraph/issues/20869

If these values differ, the pod will be evicted for exceeding its storage allotment

### Checklist

* [x] [CHANGELOG.md](https://github.com/sourcegraph/sourcegraph/blob/main/CHANGELOG.md) updated

### DIFF
--- a/base/searcher/searcher.Deployment.yaml
+++ b/base/searcher/searcher.Deployment.yaml
@@ -30,7 +30,9 @@ spec:
       - name: searcher
         env:
         - name: SEARCHER_CACHE_SIZE_MB
-          value: "50000"
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.resources.requests.ephemeral-storage
         - name: POD_NAME
           valueFrom:
             fieldRef:
--- a/base/searcher/searcher.Deployment.yaml
+++ b/base/searcher/searcher.Deployment.yaml
@@ -31,8 +31,9 @@ spec:
         env:
         - name: SEARCHER_CACHE_SIZE_MB
           valueFrom:
-            fieldRef:
-              fieldPath: spec.resources.requests.ephemeral-storage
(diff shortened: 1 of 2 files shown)


base/searcher/searcher.Deployment.yaml
```diff
@@ -30,7 +30,7 @@ spec:
       - name: searcher
         env:
         - name: SEARCHER_CACHE_SIZE_MB
-          value: "100000"
+          value: "50000"
```
keegancsmith, line 33: Looks good. Maybe we should avoid duplication here, and instead use something like

```yaml
valueFrom:
  fieldRef:
    fieldPath: spec.resources.requests.ephemeral-storage
```

That will require us updating the cache parsing code to understand kubernetes resource limit specifications, but that would be a good thing. Not for this PR, but what do you think?

keegancsmith (approved):
TIL `ephemeral-storage` limits! Finally! I see there is a whole bunch of stuff related to ephemeral-storage in k8s now, not just `emptyDir`. Would love to use `GenericEphemeralVolume` so we can get rid of the dir watcher daemonset stuff. I see that it is behind a feature flag though :(

davejrt (approved):

davejrt:
Thanks for this, I had to adjust this for a customer but never got back around to fixing the root cause.

daxmc99:
I was incorrect with my first attempt at this. This requires using `resourceFieldRef` to specify the value. This is only available in Kubernetes versions >= 1.18. 
I'm going to also upgrade our min k8s version as part of this. GKE stable is now on `1.18` so this doesn't seem super aggressive.

PR #3124, opened 2021-06-08
update chmod

davejrt:
This PR ensures that both postgres data directories have the correct permissions before the upgrade starts. There is an edge where the pgdata-11 directory could previously be remounted with the incorrect permissions, and they upgrade could not be performed.

Example error message:
```
2021-01-15 17:11:36.428 UTC [83] FATAL: data directory “/var/lib/postgresql/data” has invalid permissions
2021-01-15 17:11:36.428 UTC [83] DETAIL: Permissions should be u=rwx (0700) or u=rwx,g=rx (0750).
```

### DIFF
--- a/base/pgsql/pgsql.Deployment.yaml
+++ b/base/pgsql/pgsql.Deployment.yaml
@@ -67,7 +67,7 @@ spec:
           value: postgres://sg:@localhost:5432/?sslmode=disable
         - name: PG_EXPORTER_EXTEND_QUERY_PATH
           value: /config/queries.yaml
-        image: sourcegraph/postgres_exporter:3.27.0@sha256:4db4ff099d6aa6b6de9f3fd30407f6173509fdd8b65bfe1a24ef3a939aad4eaf
+        image: sourcegraph/postgres_exporter:3.27.1@sha256:2b1cfe5c4d08be6173c4f8277a31ca145972cc17523ab0443221bc1a874c685c
         terminationMessagePolicy: FallbackToLogsOnError
         name: pgsql-exporter
         resources:
--- a/base/pgsql/pgsql.Deployment.yaml
+++ b/base/pgsql/pgsql.Deployment.yaml
@@ -67,7 +67,7 @@ spec:
           value: postgres://sg:@localhost:5432/?sslmode=disable
         - name: PG_EXPORTER_EXTEND_QUERY_PATH
           value: /config/queries.yaml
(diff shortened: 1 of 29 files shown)


daxmc99:
This needs to be backported if we decide to do this, correct? 
Fixes https://github.com/sourcegraph/sourcegraph/issues/20425

base/pgsql/pgsql.Deployment.yaml
```diff
@@ -27,7 +27,7 @@ spec:
       initContainers:
       - name: correct-data-dir-permissions
         image: sourcegraph/alpine:3.12@sha256:ce099fbcd3cf70b338fc4cb2a4e1fa9ae847de21afdb0a849a393b87d94fb174
-        command: ["sh", "-c", "if [ -d /data/pgdata-12 ]; then chmod 750 /data/pgdata-12; fi"]
```
daxmc99, line 30: We should also do this for the https://github.com/sourcegraph/deploy-sourcegraph/pull/2902/files#diff-b822422461c5c4358eb788df612e854aa5d5701144feebb0b863c0655fe6a2f5R30

daxmc99 (changes requested):

base/pgsql/pgsql.Deployment.yaml
davejrt: > This needs to be backported if we decide to do this, correct?
> Fixes [sourcegraph/sourcegraph#20425](https://github.com/sourcegraph/sourcegraph/issues/20425)

This PR is to be merged only against 3.27...which is essentially the backport. It's the only release affected by this. The releases either side of this don't have any migration tasks. 

In my opinion, we should change this for the other db's as you correctly pointed out ( will do this shortly ) then tag the newer commit with 3.27.5. This is an infra level change that I don't think actually warrants a new patch release as there are no application level code changes. 

I recall this being discussed once before, I haven't searched for any official docs yet but will aim to do so this afternoon in order to clarify

davejrt: I don't think we need to tag a new release as all of the images are the same. We would only need to move the tag to the new commit. All of the current documentation references that tag, so there's no need update anything else.

daxmc99 (approved):
LGTM, sounds like a good idea to move the tag as part of this as well

davejrt:
closes https://github.com/sourcegraph/sourcegraph/issues/21131

PR #3149, opened 2021-06-10
user newer alpine image for overlays

davejrt:
this change uses a newer alpine image to perform updates during the conversion from root to non-root containers

[_Created by Sourcegraph batch change `dave/alpine-update`._](https://k8s.sgdev.org/users/dave/batch-changes/alpine-update)


davejrt:
fixes vulnerabilities for https://github.com/sourcegraph/security-issues/issues/151

PR #3348, opened 2021-07-09
change pgsql to alpine based container

davejrt:
related to: https://github.com/sourcegraph/security-issues/issues/151


davejrt:
Merging as this is already running in dogfood and has otherwise been tested in live running systems

related PR in dogfood https://github.com/sourcegraph/deploy-sourcegraph-dogfood-k8s/pull/2162

PR #3404, opened 2021-07-16
remove non-root overlays

davejrt:
This PR removes the non-root overlay, as it does not deploy correctly without roles/clusterroles and is largely redundant given the non-privileged overlay provides a the roles we would ideally use in a secure cluster anyway. 

Fixes https://github.com/sourcegraph/sourcegraph/issues/19724


daxmc99 (commented):
Looks good! 
Do we currently have test coverage for this? 
This previously broke silently because we were not checking that every component rolled out successfully. 
Could we write a test that does this?
Happy to work with you to test this

davejrt:
@daxmc99 we currently don't...for any of our overlays. 

We have an [issue ](https://github.com/sourcegraph/sourcegraph/issues/21123)here to test upgrades for all of our overlays. I think we should properly define the scope of that issue to include testing before the upgrade and after. 

Thoug

davejrt:
Merging this as is, with the intention that any testing be done as part of the previously mentioned issue. That's a high priority in the backlog so will get pulled into the next sprint.

PR #3499, opened 2021-07-27
change all alpine image to new image name

davejrt:
updates all remaining alpine images to new image name


daxmc99:
Why are we doing this?

davejrt:
our alpine image is now called `alpine3-12`. The older version had security vulnerabilities that were blocking salesforce https://github.com/sourcegraph/security-issues/issues/151#issuecomment-887073279

PR #3744, opened 2021-08-20
frontend: Remove SRC_GIT_SERVERS env var

tsenart:
In Kubernetes deployments, setting the SRC_GIT_SERVERS environment variable explicitly is no longer needed. Addresses of the gitserver pods will be discovered automatically. Unset it in your frontend.Deployment.yaml to make use of this feature. https://github.com/sourcegraph/sourcegraph/pull/24094

### DIFF
--- a/base/frontend/sourcegraph-frontend.Deployment.yaml
+++ b/base/frontend/sourcegraph-frontend.Deployment.yaml
@@ -54,8 +54,6 @@ spec:
           value: disable
         - name: CODEINTEL_PGUSER
           value: sg
-        - name: SRC_GIT_SERVERS
-          value: gitserver-0.gitserver:3178
         # POD_NAME is used by CACHE_DIR
         - name: POD_NAME
           valueFrom:


keegancsmith (approved):
LGTM. Maybe copy paste your changelog entry into the commit message explaining more so when downstream users get a merge conflict they have more info.

Meta: this repo has so many bot commits! surely that makes it more difficult for people who fork this repo to understand what actually changed.

tsenart:
@daxmc99: What's the status of the work you were doing to unblock this?

daxmc99 (changes requested):
This seems to be a larger issue. I thought that the only the privileged version was failing. That isn't the case. This fails for privileged & non-privileged versions after a single codehost is added. Only our privileged test adds a codehost.

keegancsmith:
Does privilaged mean something with RBAC?

tsenart:
I'm pretty sure the normal integration tests weren't failing when I first opened this PR, so if they're failing now, that's likely unrelated to the original failure reason of the RBAC failure. What have you learned about it so far in your investigation @daxmc99?

tsenart:
Here's a link to a job where the non privileged integration tests passed: https://buildkite.com/sourcegraph/deploy-sourcegraph/builds/21938

daxmc99:
Root cause: The images in this PR were out-of-date since no image update PRs had been merged in awhile. 

Merged https://github.com/sourcegraph/deploy-sourcegraph/pull/3795 to update the images, rebased this PR and all the tests passed. 

> "Why did some of the tests pass previously but not the non-privileged test"?

Non-privileged is one of our more extensive tests where we add a code host and then attempt a search.

Removing this env var has no effect UNTIL we add a code host which is why it passed the basic cluster smoke tests. 

Adding some issues to improve our test coverage so that this doesn't happen again. The RBAC again was a complete dead end but I thought it made sense do to the test that wasn't passing.

PR #4011, opened 2021-09-28
go mod tidy

davejrt:


bobheadxi (approved):
big diff haha - maybe there's a renovate option to get it to do a tidy when doing an update?

PR #4012, opened 2021-10-04
overlays: create low-resource overlay

### LINKED ISSUE #25351: deploy-sourcegraph: kustomize overlay for low-resource environments
This would be helpful to anyone wanting to deploy to a minimally-provisioned environment, such as `deploy-sourcegraph` tests as well as the `cluster-ci` tests in this repository.

daxmc99:
Closes https://github.com/sourcegraph/sourcegraph/issues/25351

### Checklist


davejrt:
part of work for https://github.com/sourcegraph/sourcegraph/issues/25977

overlays/ci/LimitRange.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: LimitRange
```
bobheadxi, line 2: wow TIL this is perfect

overlays/ci/delete-resources-2.yaml
```diff
@@ -0,0 +1,4 @@
+- op: remove
+  path: /spec/template/spec/containers/0/resources
```
bobheadxi, line 2: How is this (and the other `delete-resources`) used?

bobheadxi (approved):
Took a quick look, left a question but otherwise LGTM! Another thought: should we name this `low-resource` or something more generic?

overlays/ci/delete-resources-2.yaml
davejrt: Good catch! 

I've removed the resources all deployments now to try and make this as lean as possible.

bobheadxi:
Thanks for bringing this over the finish line @davejrt !

bobheadxi:
Non-blocking, but added delivery for post-merge review as an FYI

overlays/low-resource/README.md
```diff
@@ -0,0 +1,5 @@
+# Low Resources overlay
```
caugustus-sourcegraph, line 1: From @bobheadxi 
> should we name this low-resource or something more generic?

The generic name feels a bit misleading considering we have buildkite-specific modifications in here. Not need to change it back since this README does explain the usage. Just thinking a customer would see this and think "hey I want to use fewer resources too!"

caugustus-sourcegraph (approved):
🎉

overlays/low-resource/README.md
bobheadxi: Ah, good catch - I didn't realize there were buildkite-specific volume mounts here. I wonder if they have to be named `buildkite` and `ci`, is this specific to how agents are set up?

daxmc99:
Thank you all for merging this! 
For smoke testing clusters, I'm not sure if we want the volume mounts. 
Cons:
1. They force us  to use a certain node type with Local SSDs ($$)
1. Cleanup needs to be handled by a cron-job or some other app

Pro:
1. Faster since no waiting for google to create PVCs everytime (maybe asking for smaller SSD would be faster)
1. Higher perf

PR #4017, opened 2021-10-20
update ingress 

davejrt:
This PR updates the ingress resource api version as per [this](https://kubernetes.io/blog/2021/07/26/update-with-ingress-nginx/) notice. 

In addition, the upstream ingress-controller we use is updated. We don't install directly from upstream so customers can check-in change changes they need to make to this. 

In addition I've removed any annotations from the upstream yaml that referenced helm so this doesn't conflict with any customers environments, and so end users are not confused about this potentially needing to be deployed by helm. 

Fixes https://github.com/sourcegraph/sourcegraph/issues/16808


tests/integration/fresh/fresh_test.go
```diff
@@ -21,7 +21,7 @@ func TestFreshDeployment(t *testing.T) {
 	noCleanup := os.Getenv("NOCLEANUP") == "true"
 
 	// Validate on various kubernetes versions
-	for _, k8sVersion := range []string{"1.19", "1.20", "1.21"} {
+	for _, k8sVersion := range []string{"1.19"} {
```
daxmc99, line 24: Temp change for testing?

davejrt: Yeah, just to cut down on the churn whilst testing.

PR #4030, opened 2021-11-30
remove query-runner

camdencheek:
This removes query-runner from the sourcegraph deployment. 

### Checklist

* [x] [CHANGELOG.md](https://github.com/sourcegraph/sourcegraph/blob/main/CHANGELOG.md) updated
* [x] [K8s Upgrade notes updated](https://github.com/sourcegraph/sourcegraph/blob/main/doc/admin/updates/kubernetes.md): I don't _think_ this should require any upgrade notes, but I seem to remember deleting deployments in k8s is weird, so I don't know if our upgrade process will cause the query runner deployment to be deleted if the config just no longer exists. Could someone confirm or help me verify? Update: [PR here](https://github.com/sourcegraph/sourcegraph/pull/28378)
* [x] Sister [deploy-sourcegraph-docker](https://github.com/sourcegraph/deploy-sourcegraph-docker) change: https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/649


caugustus-sourcegraph:
> I don't know if our upgrade process will cause the query runner deployment to be deleted

@camdencheek  - Query runner will be deleted if customers run the recommended `kubectl-apply-all.sh` script, but not everyone uses that, and it's possible they'll have other monitoring that gets tripped by the deletion. It makes sense to add this to the upgrade notes for customers to be aware of / delete manually if necessary.

caugustus-sourcegraph (approved):

camdencheek:
@caugustus-sourcegraph Created a PR with the note [here](https://github.com/sourcegraph/sourcegraph/pull/28378) -- could you take a peek to see if the wording looks right?

PR #4047, opened 2021-12-23
Use default container annotations for multi-container pods

filiphaftek:
**Kubernetes deployment improvement**
This is small improvement for Pod with multi-containers, which allows to use `kubectl logs` and `kubectl exec` without specifying flag `-c`.
[Documentation](https://kubernetes.io/docs/reference/labels-annotations-taints/#kubectl-kubernetes-io-default-container)

### Checklist

## Test plan

kubectl logs ANY-POD


danieldides (approved):

daxmc99:
No changelog update needed for this

daxmc99 (approved):

base/codeintel-db/codeintel-db.Deployment.yaml
```diff
@@ -20,6 +20,8 @@ spec:
     type: Recreate
   template:
     metadata:
+      annotations:
+        kubectl.kubernetes.io/default-container: pgsql
```
caugustus-sourcegraph, line 24: Ah, looks like we had added these before but in the wrong place. Should the annotation be removed from [L6](https://github.com/sourcegraph/deploy-sourcegraph/pull/4047/files#diff-b822422461c5c4358eb788df612e854aa5d5701144feebb0b863c0655fe6a2f5L6)? (and same for all the other deployments)

PR #4050, opened 2021-12-28
Remove unused backend service

caugustus-sourcegraph:
This backend service [has existed since 2018](https://sourcegraph.com/github.com/sourcegraph/deploy-sourcegraph/-/compare/5f2ae531448c69dad0aa0721f6eb053bfd750e0b...2d9c5ce6ed578f09619d8c1fe62b53b8783574cb#diff-79a48b91330e033a0043486bb7210a85R1) without any significant change. It claims to help prevent backend pods from being scheduled on the same node, but I can't find any evidence that this actually happens. It also doesn't match with my understanding of the Kubernetes scheduler - we should be using anti-affinity rules instead now (maybe that wasn't available at the time).

Tested by spinning up a new cluster and seeing that pods that match the `-l group=backend` filter are assigned to the same node even with the backend service present. I deleted the backend service, then deleted pods in that filter several times to make sure they still distribute across nodes.

### Checklist


daxmc99 (approved):
LGTM, I was suspicious of this claim for awhile but never tested it. 
The correct way to do this in the future is with [Node Affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/) so that the frontend pods prefer to not sit on the same node

PR #4052, opened 2022-01-04
use new_migrations envvar for regression tests

davejrt:


bobheadxi (approved):
LGTM in principle, but CI seems stuck - approving to unblock when that's working :)

Linking for context, this change is a suggestion to potentially improve the stability of migrations: https://github.com/sourcegraph/sourcegraph/pull/29196#issuecomment-1005138570

PR #4055, opened 2022-01-06
Use TCP for protocol for gitserver

daxmc99:
### Checklist


caugustus-sourcegraph (approved):
This is the default protocol, so pretty safe to add. Was it causing a problem not to have it specified or is this change just for clarity?

daxmc99:
Just making the linter happy, noticed that we didn't have this set

PR #4058, opened 2022-01-14
migrator: add migrator job and require it to run

danieldides:
https://github.com/sourcegraph/sourcegraph/issues/25257 and https://github.com/sourcegraph/sourcegraph/issues/25256

This is the `migrator` job for the generic customer use case where `pgsql`, `codeintel-db`, and `codeinsights-db` are all running in cluster.

### DIFF
--- a/base/frontend/sourcegraph-frontend.Deployment.yaml
+++ b/base/frontend/sourcegraph-frontend.Deployment.yaml
@@ -27,9 +27,36 @@ spec:
         app: sourcegraph-frontend
         deploy: sourcegraph
     spec:
+      initContainers:
+      - name: migrator
+        image: index.docker.io/sourcegraph/migrator:3.36.3@sha256:4339ced184eb228da06b5050da11794ea221f47be8a7ca24369a54a8cece923b
+        args: ["up"]
+        env:
+        - name: PGDATABASE
+          value: sg
+        - name: PGHOST
+          value: pgsql
+        - name: PGPORT
+          value: "5432"
+        - name: PGSSLMODE
+          value: disable
+        - name: PGUSER
+          value: sg
+        - name: CODEINSIGHTS_PGDATASOURCE
+          value: postgres://postgres:password@codeinsights-db:5432/postgres
+        - name: CODEINTEL_PGDATABASE
+          value: sg
+        - name: CODEINTEL_PGHOST
+          value: codeintel-db
+        - name: CODEINTEL_PGPORT
+          value: "5432"
+        - name: CODEINTEL_PGSSLMODE
+          value: disable
+        - name: CODEINTEL_PGUSER
+          value: sg
       containers:
       - name: frontend
-        image: index.docker.io/sourcegraph/frontend:insiders@sha256:f6b962027b6cdb4a1cdea2573d9dfd89f353e2efb5ac2c485bf086bfa01e7a93
+        image: index.docker.io/sourcegraph/frontend:insiders@sha256:5b39d9736d201df9547391c86b08fa43ba4f500eac3c151d4dd253045aeded47
         args:
         - serve
         env:
@@ -105,7 +132,7 @@ spec:
         - mountPath: /mnt/cache
           name: cache-ssd
       - name: jaeger-agent
-        image: index.docker.io/sourcegraph/jaeger-agent:insiders@sha256:7f89e86a138d1f9a6967fb7ec04a1ea6464d1c417265445c8e5858e87616e012
+        image: index.docker.io/sourcegraph/jaeger-agent:insiders@sha256:75a3a365a89dc7377583df1e99b1955214d00dbf5d1e7d2626371733facdbcbd
         env:
           - name: POD_NAME
             valueFrom:
--- a/kubectl-apply-all.sh
+++ b/kubectl-apply-all.sh
@@ -12,5 +12,4 @@
 #
 # Apply the base Soucegraph deployment
 # shellcheck disable=SC2068
-
 kubectl apply --prune -l deploy=sourcegraph -f base --recursive $@
--- a/kubectl-apply-all.sh
+++ b/kubectl-apply-all.sh
@@ -12,4 +12,5 @@
 #
 # Apply the base Soucegraph deployment
 # shellcheck disable=SC2068
+
 kubectl apply --prune -l deploy=sourcegraph -f base --recursive $@
--- a/overlays/migrate-to-nonprivileged/frontend/sourcegraph-frontend.Deployment.yaml
+++ b/overlays/migrate-to-nonprivileged/frontend/sourcegraph-frontend.Deployment.yaml
@@ -14,3 +14,4 @@ spec:
             name: cache-ssd
           securityContext:
             runAsUser: 0
+        - name: migrator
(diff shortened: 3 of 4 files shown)


efritz (approved):
This is how clean this PR looks:

image

base/migrator/migrator.Job.yaml
```diff
@@ -0,0 +1,56 @@
...
+          - name: CODEINSIGHTS_PGUSER
+            value: "postgres"
+          - name: CODEINSIGHTS_PGPASSWORD
+            value: "password"
+          - name: CODEINSIGHTS_PGDATABASE
+            value: "postgres"
+          - name: CODEINSIGHTS_PGSSLMODE
+            value: "disable"
+      restartPolicy: OnFailure
+  backoffLimit: 0
```
caugustus-sourcegraph, line 56: What's the reasoning behind this backoffLimit? If a failure happens, it will need manual intervention to resolve even if it's a temporary problem - might not be a problem, just curious.

kubectl-apply-all.sh
```diff
@@ -12,4 +12,10 @@
 #
 # Apply the base Soucegraph deployment
 # shellcheck disable=SC2068
+
+# Run the DB migration job and wait for it to complete
+kubectl delete -f base/migrator/migrator.Job.yaml --ignore-not-found=true
+kubectl apply -f base/migrator/migrator.Job.yaml
+kubectl wait -f base/migrator/migrator.job.yaml --for=condition=complete
```
caugustus-sourcegraph, line 19: How does this command behave when the migration job fails, or takes longer than the default timeout (30s)?

base/migrator/migrator.Job.yaml
caugustus-sourcegraph: Actually, it could be a big problem when we're talking about setting up a new Sourcegraph deployment. The first time any of this runs it could be several minutes before the databases are available, and in the meantime I assume this job will fail because it can't connect to them. At that point the customer will have a broken Sourcegraph and will have to re-run the kubectl-apply-all to trigger the job again.

Since Kubernetes doesn't have the idea of dependencies I'm not sure what a good option would be. An initContainer that waits until those connections are healthy, perhaps?

base/migrator/migrator.Job.yaml
```diff
@@ -0,0 +1,56 @@
+---
```
caugustus-sourcegraph, line 1: You'll need to add this file to the kustomization bases list [here](https://sourcegraph.com/github.com/sourcegraph/deploy-sourcegraph/-/blob/overlays/bases/deployments/kustomization.yaml).

There might be some additions to the kustomize overlays too - the only one that stands out is on the non-privileged overlay, something similar to https://sourcegraph.com/github.com/sourcegraph/deploy-sourcegraph/-/blob/overlays/non-privileged/frontend/sourcegraph-frontend.Deployment.yaml so this can run as non-root.

danieldides:
Running the `migrator` as an `initContainer` to the the `frontend` seems to work as expected:

```
% k logs sourcegraph-frontend-599744cf5c-zt8qh -c migrator
t=2022-01-18T23:29:44+0000 lvl=info msg="Checked current version" schema=frontend version=1528395964 dirty=false
t=2022-01-18T23:29:44+0000 lvl=info msg="Checked current version" schema=codeintel version=1000000030 dirty=false
t=2022-01-18T23:29:44+0000 lvl=info msg="Checked current version" schema=codeinsights version=1000000024 dirty=false
t=2022-01-18T23:29:44+0000 lvl=info msg="Checked current version" schema=codeinsights version=1000000024 dirty=false
t=2022-01-18T23:29:44+0000 lvl=info msg="Upgrading schema" schema=codeinsights
t=2022-01-18T23:29:44+0000 lvl=info msg="Checked current version" schema=frontend version=1528395964 dirty=false
t=2022-01-18T23:29:44+0000 lvl=info msg="Upgrading schema" schema=frontend
t=2022-01-18T23:29:44+0000 lvl=info msg="Checked current version" schema=codeintel version=1000000030 dirty=false
t=2022-01-18T23:29:44+0000 lvl=info msg="Upgrading schema" schema=codeintel

% k logs sourcegraph-frontend-599744cf5c-zt8qh -c frontend
t=2022-01-18T23:29:45+0000 lvl=info msg="Checked current version" schema=frontend version=1528395964 dirty=false
```

danieldides:
@caugustus-sourcegraph / @daxmc99 / @kevinwojo -- Implemented the `initContainer` migrator as well as updated the `Job` to have a backoff of 4. Also updated the `kubectl-apply-all.sh` script to have a timeout of "-1" which makes it wait it's maximum (1 week).

base/frontend/sourcegraph-frontend.Deployment.yaml
```diff
@@ -27,49 +27,27 @@ spec:
...
-        # archives of repositories at a commit.
-        - name: CACHE_DIR
-          value: /mnt/cache/$(POD_NAME)
-        - name: GRAFANA_SERVER_URL
-          value: http://grafana:30070
-        - name: JAEGER_SERVER_URL
-          value: http://jaeger-query:16686
-        - name: PROMETHEUS_URL
-          value: http://prometheus:30090
+        envFrom:
```
filiphaftek, line 48: This change ca be risky, as for now, every change in ENVs causes new rollout of frontends for clients. I think that with `envFrom` this behaviour might change, as only configMap will be changed, and it does not cause rollout.

kubectl-apply-all.sh
```diff
@@ -12,4 +13,10 @@
 #
 # Apply the base Soucegraph deployment
 # shellcheck disable=SC2068
+
+# Run the DB migration job and wait for it to complete
+kubectl delete -f base/migrator/migrator.Job.yaml --ignore-not-found=true
+kubectl apply -f base/migrator/migrator.Job.yaml
+kubectl wait -f base/migrator/migrator.job.yaml --for=condition=complete --timeout=-1s
```
filiphaftek, line 20: I would propose to add comment here for customers, b/c `-1s` sounds scary to me :) They should be aware this might take long - although I know that they have docs.

filiphaftek (commented):
Left 2 comment, especially testing `envFrom` would make sense, as this can change actual behaviour of customer deployments.

base/frontend/sourcegraph-frontend.ConfigMap.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion:  v1
+kind: ConfigMap
```
daxmc99, line 2: Are we keeping this change in for this PR? If so, I will add it to the overlays. Otherwise it needs to be removed and the env values added to the initContainer. (Happy to do either, just need a decision for this PR)

daxmc99 (changes requested):

danieldides:
This PR ultimately did not make it in the `3.36` release for a variety of reasons. I'm going to move this back to a `Draft` and we can re-evaluate how much of it we want to salvage for the next release. There is a separate PR (https://github.com/sourcegraph/deploy-sourcegraph/pull/4062) to move the `Job` spec to the `configure` folder that can be released and matches the current documentation (https://github.com/sourcegraph/sourcegraph/pull/29779).

caugustus-sourcegraph:
@danieldides - When do you think this PR will be ready for review? I think the migrator is getting close to being ready for final testing, so it'd be great to have these changes available to test next week.

danieldides:
@caugustus-sourcegraph I'll get these changes fixed up and ready to go this week and then we define and execute some sort of test plan next week.

PR #4062, opened 2022-01-20
Add migrator job spec to configure

danieldides:
Move Job files to `configure/` so they can be applied later if a customer wants

### DIFF
--- a/configure/migrator/migrator.Job.yaml
+++ b/configure/migrator/migrator.Job.yaml
@@ -1,0 +1,126 @@
+# These manifests are provided in the event that someone may wish to execute
+# database migrations outside of a code deployment.
+#
+# For more information, please see: https://docs.sourcegraph.com/admin/install/kubernetes/operations#database-migrations
+
+# ---
+# apiVersion: batch/v1
+# kind: Job
+# metadata:
+#   name: migrator
+#   annotations:
+#     description: Database migrator for Postgres databases
+#   labels:
+#     app.component.kubernetes/component: migrator
+#     deploy: sourcegraph
+# spec:
+#   template:
+#     spec:
+#       containers:
+#       - name: migrator
+#         image: "sourcegraph/migrator:3.36.0"
+#         args: ["up"]
+#         env:
+#           - name: PGHOST
+#             value: "pgsql"
+#           - name: PGPORT
+#             value: "5432"
+#           - name: PGUSER
+#             value: "sg"
+#           - name: PGPASSWORD
+#             value: "sg"
+#           - name: PGDATABASE
+#             value: "sg"
+#           - name: PGSSLMODE
+#             value: "disable"
+#           - name: CODEINTEL_PGHOST
+#             value: "codeintel-db"
+#           - name: CODEINTEL_PGPORT
+#             value: "5432"
+#           - name: CODEINTEL_PGUSER
+#             value: "sg"
+#           - name: CODEINTEL_PGPASSWORD
+#             value: "sg"
+#           - name: CODEINTEL_PGDATABASE
+#             value: "sg"
+#           - name: CODEINTEL_PGSSLMODE
+#             value: "disable"
+#           - name: CODEINSIGHTS_PGHOST
+#             value: "codeinsights-db"
+#           - name: CODEINSIGHTS_PGPORT
+#             value: "5432"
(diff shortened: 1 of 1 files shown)


configure/migrator/migrator.Job.yaml
```diff
@@ -0,0 +1,184 @@
+# These manifests are provided in the event that someone may wish to execute
+# database migrations outside of a code deployment.
+#
+# For more information, please see: https://docs.sourcegraph.com/admin/install/kubernetes/operations#database-migrations
+
+# ---
```
caugustus-sourcegraph, line 6: You can uncomment this job - it won't get called automatically so it's safe to leave in the "active" state.

configure/migrator/migrator.Job.yaml
```diff
@@ -0,0 +1,184 @@
...
+#     description: Database migrator for Postgres databases
+#   labels:
+#     app.component.kubernetes/component: migrator
+#     deploy: sourcegraph
+# spec:
+#   template:
+#     spec:
+#       containers:
+#       - name: migrator
+#         image: "sourcegraph/migrator:3.36.0"
```
caugustus-sourcegraph, line 21: Can you update this to follow the same pattern as our other deployments - something similar to:
`image: index.docker.io/sourcegraph/frontend:insiders@sha256:f6b962027b6cdb4a1cdea2573d9dfd89f353e2efb5ac2c485bf086bfa01e7a93`

I can't say for sure but I think we rely on that pattern to do some checks (https://sourcegraph.com/github.com/sourcegraph/deploy-sourcegraph/-/blob/tools/check-image-names/check-image-names.go?L17). Super nitpicky, sorry!

configure/migrator/migrator.Job.yaml
```diff
@@ -0,0 +1,184 @@
...
+#             value: "postgres"
+#           - name: CODEINSIGHTS_PGPASSWORD
+#             value: "password"
+#           - name: CODEINSIGHTS_PGDATABASE
+#             value: "postgres"
+#           - name: CODEINSIGHTS_PGSSLMODE
+#             value: "disable"
+#       restartPolicy: OnFailure
+#   backoffLimit: 4
+# ---
```
caugustus-sourcegraph, line 62: What are your thoughts on removing the two examples below, and just documenting that you can customize the `args` command? I'm on the fence - I do like the explicit example, but it might cause some confusion. I could easily see someone thinking they have to run all three of these instead of noticing they're options.

configure/migrator/migrator.Job.yaml
```diff
@@ -0,0 +1,184 @@
...
+#     app.component.kubernetes/component: migrator
+#     deploy: sourcegraph
+# spec:
+#   template:
+#     spec:
+#       containers:
+#       - name: migrator
+#         image: "sourcegraph/migrator:3.35.1"
+#         args: ["up", "-db", "codeinsights"]
+#         env:
```
caugustus-sourcegraph, line 145: If we do end up keeping these examples, this `env` list can be trimmed down to just CodeInsights related stuff.

danieldides: There's currently an issue requiring having all the environments configured to be able to migrate `codeinsights-db`: https://github.com/sourcegraph/sourcegraph/issues/30013

caugustus-sourcegraph: Nice find!

configure/migrator/migrator.Job.yaml
danieldides: I think since they're all commented this is a fine solution. I can push a new commit with all the environment variables (to cover the `codeinsights-db` issue) and then we can update the documentation to reflect how to modify the command.

configure/migrator/migrator.Job.yaml
danieldides: I just pushed the change with the digest pattern and the correct 3.36.1 image

https://hub.docker.com/layers/sourcegraph/migrator/3.36.1/images/sha256-54a1b8db471b0d13d02bf730551cb748d8362cf3bb35c09f20aa02b8c918c56d?context=explore

configure/migrator/migrator.Job.yaml
danieldides: Done!

caugustus-sourcegraph (approved):

PR #4068, opened 2022-01-27
Run each live cluster test as a separate test for easier retries

caugustus-sourcegraph:
The integration tests create real GKE clusters to perform validation against multiple Kubernetes versions. It's very common that one cluster will fail validation, while the others succeed. Retrying the test means we have to re-run validation against cluster versions that already passed, which is slow and risks a formerly-successful version failing.

This PR makes it so that we run each Kubernetes version as a separate buildkite step, so that each can be retried individually. It is a short-term answer to reduce the pain until we can put more investment into fixing the flakiness.

Better potential solutions long-term:
- Investigate and resolve the root causes of the flaky tests (debugging the causes has proven difficult)
- Stop using GKE clusters, use kind or similar lightweight Kubernetes options
- Journal test progress so that a retry only attempts failed tests
- Abandon modern technology and live in a tent in the forest

### Checklist

### DIFF
--- a/.buildkite/integration-test.sh
+++ b/.buildkite/integration-test.sh
@@ -16,6 +16,8 @@ export BUILD_CREATOR
 GENERATED_BASE=$(mktemp -d)
 export GENERATED_BASE
 
+export TEST_K8S_VERSION=$1
+
 cleanup() {
   echo "--- Cleaning up test artifacts"
   rm -rf "${GENERATED_BASE}"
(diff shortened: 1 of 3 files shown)


caugustus-sourcegraph:
[This build](https://buildkite.com/sourcegraph/deploy-sourcegraph/builds/24726) demonstrates the experience - version 1.21 failed and was able to be retried independently.

kevinwojo (approved):
This is pretty great work. 

I know we have discussions around which versions of k8s we are supporting, so good to know this is there. :-)

Excited to have some speedup with builds!

michaellzc (approved):
🎉

PR #4069, opened 2022-01-28
Run update tag script against overlays

caugustus-sourcegraph:
The `overlays/migrate-to-nonprivileged` overlays include an image reference, which is unique among the overlays. The `update-docker-tags.sh` was only running against `base` so these overlays were getting skipped during releases and always staying on an old version.

On the fence about whether to include the actual manifest changes here or just the script itself. [Renovate is currently broken](https://github.com/sourcegraph/sourcegraph/issues/29473) in this repo so the other image references are quite out of date. I can revert if preferred.

I don't think this is worth doing a patch for - no customer has mentioned this, and it's been broken for a long time. We can just fix it for going forward.

### Checklist

### DIFF
--- a/overlays/migrate-to-nonprivileged/frontend/sourcegraph-frontend.Deployment.yaml
+++ b/overlays/migrate-to-nonprivileged/frontend/sourcegraph-frontend.Deployment.yaml
@@ -7,7 +7,7 @@ spec:
     spec:
       initContainers:
         - name: transfer-cache
-          image: index.docker.io/sourcegraph/alpine-3.12:insiders@sha256:b5716faa7c5d40542132176960ae2d5708a20a8dd5dcd1e63d41af4c27a1b0c4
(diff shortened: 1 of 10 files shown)


michaellzc (approved):
Not blocking, just a question.

It looks to me the overlay is just injecting an `intContainer` and run some `chown` to fix the ownership.

Does it matter which image we're using to run the `chown` command at all?

The only shard content between initContainera and the primary container is the `/mnt/cache` volume.

So technically, nothing is broken.

Unless there's some magic sauce baked into our custom alpine image that I didn't account for.

caugustus-sourcegraph:
@michaellzc - Ah, great points that these are transient containers so the risk is low and it's probably been working fine! The consequences aren't that bad as a result:
* The alpine images don't have any recent security fixes (they're 6+ months old at this point) but they're short-lived
* They're using an insiders build which is technically not part of any release
* If a customer is using a private registry, it's an extra image they have to mirror

Now I'm extra convinced that we don't need a patch 😄

michaellzc:
Yep, no need for a patch release!

PR #4077, opened 2022-02-04
Clarify Prometheus configuration

daxmc99:
Part of https://github.com/sourcegraph/customer/issues/679

### Test plan
n/a

### DIFF
--- a/base/prometheus/prometheus.ConfigMap.yaml
+++ b/base/prometheus/prometheus.ConfigMap.yaml
@@ -206,7 +206,7 @@ data:
     # The relabeling allows the actual pod scrape endpoint to be configured via the
     # following annotations:
     #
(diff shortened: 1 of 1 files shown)


base/prometheus/prometheus.ConfigMap.yaml
```diff
@@ -205,7 +206,7 @@ data:
     # The relabeling allows the actual pod scrape endpoint to be configured via the
     # following annotations:
     #
-    # * `prometheus.io/scrape`: Only scrape pods that have a value of `true`
+    # * `sourcegraph.prometheus.io/scrape`: Only scrape pods that have a value of `true`
```
danieldides, line 209:
```suggestion
    # * `sourcegraph.prometheus/scrape`: Only scrape pods that have a value of `true`
```

danieldides (changes requested):

PR #4086, opened 2022-02-22
Update restricted integration tests to clean up abandoned PVC’s

caugustus-sourcegraph:
Partially addresses https://github.com/sourcegraph/sourcegraph/issues/31262 by eliminating the main culprit.

When running restricted tests, we delete the cluster without deleting the resources first, which causes abandoned PVC's. Some days we generate hundreds of PVC's (I cleared out the disks from CI on Monday morning, and today there are 205 new ones). This test is the only one that behaves this way.

The abandoned-disks issue seems to be a common complaint without any real solution - the [official bug report](https://issuetracker.google.com/issues/121034250?pli=1) recommends deleting the namespace first, but that won't always fix it. As a workaround, I've added both the namespace deletion ~~and a step to try to delete abandoned disks through gcloud.~~ (removed this action because it doesn't belong in a test)

This is an incomplete fix since it only catches disks created through this test; a separate followup will be to add monitoring across all our projects to find abandoned disks.

### Checklist

### DIFF
--- a/tests/integration/restricted/test.sh
+++ b/tests/integration/restricted/test.sh
@@ -38,7 +38,7 @@ gcloud container clusters get-credentials ${CLUSTER_NAME} --zone ${TEST_GCP_ZONE
 # Configure if the test should clean up after itself - useful for debugging
 if [ "${NOCLEANUP:-}" != "true" ]; then
   CLUSTER_CLEANUP="gcloud container clusters delete ${CLUSTER_NAME} --zone ${TEST_GCP_ZONE} --project ${TEST_GCP_PROJECT} --quiet"
-  DISK_CLEANUP="gcloud compute disks delete $(gcloud compute disks list --format="table(name,users)" --filter="name~^gke-${CLUSTER_NAME_PREFIX}.*-pvc-.* AND -users:*" --project ${TEST_GCP_PROJECT}) --zone ${TEST_GCP_ZONE} --project ${TEST_GCP_PROJECT} --quiet"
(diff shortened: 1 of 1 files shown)


tests/integration/restricted/test.sh
```diff
@@ -37,7 +38,8 @@ gcloud container clusters get-credentials ${CLUSTER_NAME} --zone ${TEST_GCP_ZONE
 # Configure if the test should clean up after itself - useful for debugging
 if [ "${NOCLEANUP:-}" != "true" ]; then
   CLUSTER_CLEANUP="gcloud container clusters delete ${CLUSTER_NAME} --zone ${TEST_GCP_ZONE} --project ${TEST_GCP_PROJECT} --quiet"
-  CLEANUP="$CLUSTER_CLEANUP; $CLEANUP"
+  DISK_CLEANUP="gcloud compute disks delete $(gcloud compute disks list --format="value(name)" --filter="name~^gke-${CLUSTER_NAME_PREFIX}.*-pvc-.* AND -users:*" --project ${TEST_GCP_PROJECT}) --zone ${TEST_GCP_ZONE} --project ${TEST_GCP_PROJECT} --quiet"
```
caugustus-sourcegraph, line 41: The -users:* filter means "don't include anything that has a user attached" - i.e., this is an in-use disk and should not be deleted. https://groups.google.com/g/gce-discussion/c/RLrwOx8fazo/m/9ve7lIdsBQAJ

One tiny pitfall with this command is that it cannot be limited to only deleting disks associated with the current test run. Disks created through the kubernetes.io/gce-pd provider do not attach any useful labels to the disk that will let us filter.

We could try switching to a new storage provisioner but in the end I don't think it matters too much. We'll only be deleting abandoned disks, in the CI project, with the gke-ds-test-restricted prefix.

caugustus-sourcegraph: Removed this change - it doesn't belong in a test since it isn't restricted to the current run, and the namespace deletion provides enough value on its own.

michaellzc (approved):
lgtm

PR #4110, opened 2022-04-07
ci: output which file is failing verify-labels

keegancsmith:
A file was failing to parse with yj. This was harder than it should of been to track down. This commit improves the verify-label script to run per file so we know which file is responsible.

Other improvements:
- Check if the value is sourcegraph, not containing sourcegraph. Uses jq equality check.
- Don't install yj if already present. Useful for testing locally.
- Use bash on env. bash at /bin can be old (and doesn't exist on NixOS).

Test Plan: modified some yaml files to fail the rule and ran .buildkite/verify-label.sh.

### DIFF
--- a/.buildkite/verify-label.sh
+++ b/.buildkite/verify-label.sh
@@ -1,15 +1,14 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
-set -ex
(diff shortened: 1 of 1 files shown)


filiphaftek (approved):
👍

PR #4115, opened 2022-04-13
Add init container to codeinsights-db to fix data dir permissions

caugustus-sourcegraph:
Added the same initContainer that is present for codeintel-db and pgsql to the codeinsights-db deployment.

## Context

When migrating from timescale to postgres on the non-privileged overlay in a GKE cluster, the following error shows up:

```
codeinsights-db-59fddf6dbf-wlgfs codeinsights waiting for server to start....2022-04-12 18:52:50.561 UTC [16] FATAL:  data directory "/var/lib/postgresql/data/pgdata" has invalid permissions
```

And it's true! That directory has the following permissions set at startup (before postgres starts), even on the timescale docker image:
```
Access: (2770/drwxrws---)  Uid: (   70/postgres)   Gid: (   70/postgres)
```

By the time postgres starts, the permissions are rewritten to be correct:
```
Access: (0700/drwx------)  Uid: (   70/postgres)   Gid: (   70/postgres)
```

This error doesn't happen when upgrading the base deployment, or when starting a new non-privileged deployment. Strangely, I also can't make it happen in my `kind` cluster - only when running against a GKE cluster. We also saw this behavior when switching dogfood to helm, although I can't remember if it was also using the new postgres image.

I haven't been able to track down why the behavior is different between the timescale image and postgres, or why it's only a factor in GKE. I figure it's safer to just go ahead and run the initContainer to be consistent with the other databases, and we can dig in more fully later on.

### Checklist

### Test plan

Upgrade from 3.38 to latest, using the restricted overlay. Code insights does not start before this change; starts successfully afterwards and $PGDATA ends up the the `0700` permissions (so some kind of permission rewriting is happening during the postgres startup).

### DIFF
--- a/base/codeinsights-db/codeinsights-db.Deployment.yaml
+++ b/base/codeinsights-db/codeinsights-db.Deployment.yaml
@@ -24,6 +24,22 @@ spec:
         app: codeinsights-db
         group: backend
     spec:
+      initContainers:
+      - name: correct-data-dir-permissions
+        image: index.docker.io/sourcegraph/alpine-3.12:insiders@sha256:3d926ffec4cdd54b5197bf16dded63e37b1a7045e2623ae9f8eda0a63bbf8231
+        command: ["sh", "-c", "if [ -d /var/lib/postgresql/data/pgdata ]; then chmod 750 /var/lib/postgresql/data/pgdata; fi"]
+        volumeMounts:
+        - mountPath: /var/lib/postgresql/data/
+          name: disk
+        securityContext:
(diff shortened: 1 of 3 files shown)


michaellzc (approved):
with dogfood
1. migrated from timescale to plain postgres
1. then we do the non-privileged migration
1. then we move to helm

things went wrong between step 2 and 3 but then it magically got fixed overnight (????)

🤔 maybe it's some weird behaviour with GKE PD?

PR #4123, opened 2022-04-22
fix: make migrator flags consistent across deployments

kevinwojo:
A fix that corrects the default behavior of the `migrator` service is included in this release. An attempt to standardize CLI packages in v3.39.0 unintentionally broke the default behavior. In order to guard against this, all command line arguments are explicitly set in the deployment manifest.

https://github.com/sourcegraph/customer/issues/845

### Checklist

* [x] [K8s Upgrade notes updated](https://github.com/sourcegraph/sourcegraph/blob/main/doc/admin/updates/kubernetes.md)

### DIFF
--- a/configure/migrator/migrator.Job.yaml
+++ b/configure/migrator/migrator.Job.yaml
@@ -19,7 +19,7 @@ spec:
       containers:
       - name: migrator
         image: "index.docker.io/sourcegraph/migrator:3.38.1@sha256:4bd4fd68fc3e6d1c947847aec126ff8c4577da80cbd11252c6c72a482722ad66"
-        args: ["up"]
+        args: ["up", "-db=all"]
         env:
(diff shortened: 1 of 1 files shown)


configure/migrator/migrator.Job.yaml
```diff
@@ -19,7 +19,9 @@ spec:
       containers:
       - name: migrator
         image: "index.docker.io/sourcegraph/migrator:3.38.1@sha256:4bd4fd68fc3e6d1c947847aec126ff8c4577da80cbd11252c6c72a482722ad66"
-        args: ["up"]
+        args:
+          - up
+          - -db=all
```
caugustus-sourcegraph, line 24: I'm undecided about whether we need to make this change, but if we do, shouldn't the frontend deployment also be updated?

PR #4140, opened 2022-06-10
IGNORE ME

efritz:
NOT TO MERGE ONLY TO SHOW


efritz:
Full instructions to reproduce:

- Move `base/executors` directory out of clone momentarily. Run `kubectl apply --prune -l deploy=sourcegraph -f base --recursive` to start a stack without executors at first
- Forward a port to the frontend via `kubectl port-forward sourcegraph-frontend-{one of your frontend pods} :3080`
- Visit `localhost:{the port output from the previous step}` in a browser and ensure the instance is healthy before adding executors
- Go to site settings and fill out `externalURL` (documented as `EXECUTOR_FRONTEND_URL` here until the next release), add executor access tokens, and turn on auto-indexing
- Add repositories
- Run `kubectl apply -l deploy=sourcegraph -f base/executor/` to launch an executor instance, which should connect to the frontend
- Infer an auto-index job for a Go repository
- Ensure we have precise intel for HEAD on that repo

PR #4147, opened 2022-06-30
Add kustomize overlay for updating resources

### LINKED ISSUE #38029: Create a kustomize overlay  for resource estimator
For non-Helm deployments - should we recommend using a kustomize overlay instead? I'm not sure we have an existing example, though, so that's probably out of scope for this doc.

_Originally posted by @caugustus-sourcegraph in https://github.com/sourcegraph/sourcegraph/pull/37869#discussion_r908831178_

abeatrix:
Follows up from https://github.com/sourcegraph/sourcegraph/pull/37869#discussion_r908831178
Close https://github.com/sourcegraph/sourcegraph/issues/38029

This PR adds a new kustomize overlay that acts as an example for using kustomize to update resources based on the results from our resource estimator

### Checklist

^ Not applicable since the overlay does not involve images and only works in k8s deployments

### Test plan

Tested manually - files in the generated cluster were updated correctly


kevinwojo (approved):
I think this is good supporting material. Do you plan on cross-linking to the resource estimator or other docs?

abeatrix:
> I think this is good supporting material. Do you plan on cross-linking to the resource estimator or other docs?

I do! I will add it to our Kustomize docs where the resource estimator links to :)

PR #4150, opened 2022-07-12
Add envoy filter overlay

abeatrix:
### Checklist

Add a new envoy filter overlay as an example to resolve the following error for instances that are running service mesh (ex. istio):

```
Strconv.Atoi: parsing "": invalid syntax
```

This error occurs when the service mesh like istio drops trailers for http1 by default. Enable trailers using this overlay resolves the issue.  

image

### Test plan

Tested locally on a minikube instance with istio enabled

### DIFF
--- a/overlays/envoy/README.md
+++ b/overlays/envoy/README.md
@@ -1,6 +1,6 @@
 # Kustomize Overlay - Envoy Filter
 
-This overlay applies a new envoy filter to enable trailers for gitserver to resolve the following error message caused by enabling service mesh:
+This overlay applies a new envoy filter to enable HTTP trailers for gitserver to resolve the following error message caused by enabling service mesh:
 
 ```
 "git command [git rev-parse HEAD] failed (stderr: \"\"): strconv.Atoi: parsing \"\"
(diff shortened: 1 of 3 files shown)


overlays/envoy/README.md
```diff
@@ -0,0 +1,31 @@
+# Kustomize Overlay - Envoy Filter
+
+This overlay applies a new envoy filter to enable trailers for gitserver to resolve the following error message caused by enabling service mesh:
```
keegancsmith, line 3:
```suggestion
This overlay applies a new envoy filter to enable HTTP trailers for gitserver to resolve the following error message caused by enabling service mesh:
```

keegancsmith (approved):
I'm not sure who is best to review this, but I took a look and it makes sense to me. cc @gabtorre

keegancsmith:
Do we not need to do anything to resolve issues with indexed-search?

abeatrix:
> Do we not need to do anything to resolve issues with indexed-search?

I believe that won't be necessary since the error comes from [gitserver](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@v3.41.0/-/blob/internal/gitserver/git_command.go?L181-183), and I did not encounter any issue with indexed-search using this overlay 🙏

PR #4164, opened 2022-08-26
frontend: remove unused CACHE_DIR

keegancsmith:
The code backing this variable has been unused for atleast a year. Removing this also allows us to simplify deployment, since right now we tell customers to assign ephemeral storage to frontend (but that will never be used!)

Test Plan: CI. I previously did a lot of validation that we can remove this (reading code, inspecting prod and dogfood environments for cache dir usage)

Part of https://github.com/sourcegraph/sourcegraph/issues/38934


base/frontend/sourcegraph-frontend.Deployment.yaml
```diff
@@ -89,15 +89,6 @@ spec:
...
           value: sg
-        # POD_NAME is used by CACHE_DIR
-        - name: POD_NAME
-          valueFrom:
-            fieldRef:
-              fieldPath: metadata.name
-        # CACHE_DIR stores larger items we cache. Majority of it is zip
-        # archives of repositories at a commit.
-        - name: CACHE_DIR
-          value: /mnt/cache/$(POD_NAME)
```
eseliger, line 100: We should also remove the volume mount and emptydir directive for it

keegancsmith: hah I spotted that as well, just pushed the update.

eseliger (approved):
thanks!

keegancsmith:
Just noticed there are a few references to this in docs/overlays. Will update.

keegancsmith:
sorry for the ping, I think I worked it out :)

keegancsmith:
Alright I have no idea, if the test output some logs from the cluster maybe I could work it out. But a total guess is maybe we need my corresponding change in sourcegraph/sourcegraph to land first. So I will just pick this back up on Monday and hope for the best.

bobheadxi:
```
+ /usr/local/bin/src version
--
  | Current version: 3.35.1
  | json: cannot unmarshal string into Go value of type struct { Version string "json:\"version\"" }
```

This occurs right before the error, which is interesting - not something we see in master 🤔

kalanchan:
@keegancsmith changes look good, but just to confirm this is for customer deployments? changes to dotcom are in https://github.com/sourcegraph/deploy-sourcegraph-cloud. We should get @sourcegraph/delivery in here to double check as well

eseliger:
This was fixed on latest main!

keegancsmith:
@kalanchan this is for every deploy. I introduced this variable back in the day (including to this repo), I know we can remove it :)

PR #4167, opened 2022-08-29
restricted integration: run src version against local version

keegancsmith:
It was running against sourcegraph.com, so was a pointless src-cli call.

Test Plan: CI is green


burmudar (approved):

eseliger (commented):
Good catch!

PR #4198, opened 2022-11-02
Implement kustomization files to base

abeatrix:
### Checklist

### Test plan

Close https://github.com/sourcegraph/customer/issues/693 & https://github.com/sourcegraph/sourcegraph/issues/43831

The changes made in this PR is supposed to work with the new [sourcegraph/kustomize](https://github.com/sourcegraph/kustomize) repo. 

By adding `kustomization.yaml` files directly to the base layer, users will no longer be required to clone this reference repository directly onto their machine to use kustomize. This is because [kustomize supports remote build](https://sourcegraph.com/github.com/kubernetes-sigs/kustomize@master/-/blob/examples/remoteBuild.md), where user can just reference this repository in their kustomization file ([example](https://github.com/sourcegraph/kustomize/blob/75a76bee1731a02fdf331caac9d88e599c38d0bd/overlays/non-privileged/kustomization.yaml)) through its URL.

Eliminating the needs of cloning this repository also prevents users from making changes directly to the base files.

## Test

To test the changes made in this PR:
1. create a new folder anywhere on your local machine. Name the folder `overlays`
2. Create a new file called `kustomization.yaml` inside the `overlays` folder
3. Copy and paste the code below to the `kustomization.yaml` file
4. `cd` back to one level before the `overlays` folder and run `kustomize build overlays`
5. you should see a new set of kustomized yaml files in your terminal output

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
namespace: ns-sourcegraph
resources:
  - git::git@github.com:sourcegraph/deploy-sourcegraph/base?ref=bee/kustomize
components:
  - git::git@github.com:sourcegraph/kustomize/components/non-privileged
  - git::git@github.com:sourcegraph/kustomize/components/non-privileged-create-cluster
```


daxmc99:
For users currently using our `kubectl-apply.sh` script in the repo base, does this break their workflow?

unknwon:
Not a review comment: Looks like every YAML file is missing a new line EOF.

abeatrix:
> For users currently using our `kubectl-apply.sh` script in the repo base, does this break their workflow? 

It shouldn't if they are using a newer version of kubectl. It's working fine for me locally but it looks like CI is having issue with that? Changing "kubectl apply -f" to "kubectl apply -k" should work though 🤔

michaellzc:
lgtm

regarding the apply script, it should just work for customers provided they never customize it, so it's worth pinging CE and application engineers for a quick check.

notes about the use of `kubectl -k`, I dislike the use of `-k` flag because it doesn't work all the time, e.g. you cannot run `kubetl diff -k`. It's always best to run

```sh
kustomize build  ./path/to/overlay/. | kubectl diff -f -
kustomize build  ./path/to/overlay/. | kubectl apply -f -
```

this way is a lot more robust ;)

michaellzc:
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
namespace: ns-sourcegraph
resources:
  - git::git@github.com:sourcegraph/deploy-sourcegraph/base?ref=bee/kustomize
components:
  - git::git@github.com:sourcegraph/kustomize/components/non-privileged
  - git::git@github.com:sourcegraph/kustomize/components/non-privileged-create-cluster
```

🤔 maybe just use `https`? this should work

`https://github.com/sourcegraph/deploy-sourcegraph//base?ref=bee/kustomize`

michaellzc:
is there anything specific an existing customer has to watch out for during the migration? 
would love to see something in the upcoming docs!
https://github.com/sourcegraph/sourcegraph/issues/43837

abeatrix:
As discussed with Michael, migration is currently not supported and the new overlays are  intended for new deployments only. 

The old version will be marked as deprecated but wont be removed from the repo yet in case of deployments that are depending on them. 

Currently working on the docs to reflect this new update.

burmudar:
@abeatrix `kubectl` was previously complaining about not know the `-` option so I fixed it but now it's complaining about something else
```
  | + kustomize build ./configure/.
  | + kubectl apply --prune -l deploy=sourcegraph --dry-run --validate --context=gke_sourcegraph-dev_us-central1-f_cloud -f -
  | Error: unable to find one of 'kustomization.yaml', 'kustomization.yml' or 'Kustomization' in directory '/root/buildkite/build/deploy-sourcegraph/configure'
  | W1111 08:42:30.471149    7359 helpers.go:557] --dry-run is deprecated and can be replaced with --dry-run=client.
  | error: no objects passed to apply
  | 🚨 Error: The command exited with status 1
  | user command error: exit status 1

```

.buildkite/verify-yaml.sh
```diff
@@ -5,7 +5,7 @@ cd "$(dirname "${BASH_SOURCE[0]}")/.."
 
 gcloud container clusters get-credentials cloud --zone us-central1-f --project sourcegraph-dev
 
-kubectl apply --dry-run --validate --recursive -f base/ --context=gke_sourcegraph-dev_us-central1-f_cloud
-kubectl apply --dry-run --validate --recursive -f configure/ --context=gke_sourcegraph-dev_us-central1-f_cloud
+kustomize build ./base/. | kubectl apply --dry-run --validate --context=gke_sourcegraph-dev_us-central1-f_cloud -f -
+kustomize build ./configure/. | kubectl apply --prune -l deploy=sourcegraph --dry-run --validate --context=gke_sourcegraph-dev_us-central1-f_cloud -f -
```
burmudar, line 9: `./configure/.` doesn't contain any `kustomize.yaml` files which is why this line is failing on the pipeline

abeatrix: Hey will, thanks for looking into this! I will fix the rest of the errors about the kustomize files tomorrow. 

As for the gloud auth plugin error i mentioned in the original thread, you can find it in the commit I made before today. (So sorry for the confusions cause since I haven't checked the build after the commits I made today lol)

michaellzc (approved):
lgtm

delegate final sign-off for CE working with customers that are using kustomize

abeatrix:
Moved to https://github.com/sourcegraph/deploy-sourcegraph-k8s

PR #4205, opened 2022-12-04
make cadvisor unprivileged, remove PodSecurityPolicy 

beyang:
make cadvisor unprivileged, remove PodSecurityPolicy (removed in latest k8s)

### Checklist

k8s 1.25 removes PodSecurityPolicy, so we need to purge these from our manifest. The only reference is to cadvisor's PodSecurityPolicy. Once this is removed, the ClusterRole and ClusterRoleBinding become unnecessary, so this PR removes them.

This change does not require a corresponding change to the Docker Compose config, since PodSecurityPolicy is a k8s-specific topic. There *is* a sister PR in deploy-sourcegraph-helm, which removes the ClusterRole and ClusterRoleBinding if the PodSecurityPolicy option is not explicitly enabled for cadvisor in the chart.

### Test plan

Tested against a GCP k8s cluster to verify neither cadvisor nor pods in the rest of the app break.


slimsag (approved):
Epic!

PR #4207, opened 2022-12-06
demonstration: sg ops update not working

slimsag:
DO NOT MERGE

This is a demonstration that `sg ops update` doesn't really work today.

This change was produced via:

```
sg ops update-images -t 187572_2022-12-06_cbecc5321c7d -kind k8s ./base
```

Most changes here are innocuous, but a few formatting changes here are quite aggressive:


slimsag:
This appears to produce the least intrusive diff (still formatting in some files, but not too bad):

```
sg ops update-images -t 187572_2022-12-06_cbecc5321c7d -kind k8s ./base
git checkout -- **/*.ConfigMap.yaml **/*.Role.yaml **/*.Ingress.yaml **/*.Service.yaml **/*.IndexerService.yaml
git add base/
```

slimsag:
Another complaint/observation:

`sg ops update-images` cannot update the `pure-docker/` directory in https://github.com/sourcegraph/deploy-sourcegraph-docker which means the changes it wants to make to that repository are always incomplete (it would be bad to merge a change there to only Docker Compose and not the bash scripts.)

PR #4215, opened 2022-12-15
feat: non-privileged base cluster with new kustomize

abeatrix:
# New default cluster with improved deployment setup

❌ DO NOT MERGE UNTIL https://github.com/sourcegraph/deploy-sourcegraph/pull/4215 IS CLOSED ❌ 

RE: https://github.com/sourcegraph/deploy-sourcegraph/pull/4213,  [job fair item](https://docs.google.com/document/d/1pR29WO-D883QvhCQJlnOv_Qu5W59jf0ktWNhdYJd08c/edit#)

Related PRs: https://github.com/sourcegraph/sourcegraph/pull/46227 https://github.com/sourcegraph/sourcegraph/pull/45862

### Introduction

This PR serves to introduce a new base cluster for Sourcegraph, along with a new Kustomize setup for deploying Sourcegraph that aligns with industry standards and provides a better user experience.

### Background

The current base cluster includes all resources that may not be suitable for all user environments, resulting in permission issues during deployments and causing delays and a poor customer experience. In an effort to address these issues, we have developed a new base cluster that is tailored to the scaling and security requirements of enterprise customer environments.

### New base cluster

The new base cluster is made up of two different stacks:

- Sourcegraph Main Stack (base) includes services that are required to run the Sourcegraph main application.
  - Default to run without RBAC resources
- Sourcegraph Monitoring Stack (monitoring) includes services that are required to run the monitoring services tailored for Sourcegraph.
  - Default to run with RBAC resources
  - Can be deployed separately 

#### Notable changes and improvements

- Default services to run with non-root users
- Default services to run with non-privileged users
- Default resources tested for Size XS instance
- Separate services into "base" and "monitoring" stacks
- Remove all RBAC resources from the main (base) application
- Relocate frontend env vars into a ConfigMap for better visibility
- Work with Kustomize without additional setup or tools
- Introduce new env vars documented in https://github.com/sourcegraph/sourcegraph/pull/45862

### New Kustomize setup

The new setup introduces a better deployment experience for everyone, while addressing the following common issues caused by our current Kubernetes deployment method:

- Not applicable to everyone
- Lack of user-friendliness and beginner-friendliness
- Lack of flexibility to remove resources or services
- Dependence on custom scripts that are not a part of traditional methods
- Confusion caused by the use of non-standard custom scripts, result in broken instances
- Dependence on users reading docs

#### Benefits

The changes we are introducing provide users with higher flexibility to customize their Sourcegraph deployment using the components that are pre-configured by Sourcegraph. This in turn provides us with the ability to control and oversee the changes customers make and prevent incidents where clusters are set up with service-breaking configurations.

It also allows different services to be deployed separately. No one should be blocked due to permission issue 

### Expectations

The new changes should**NOT** affect current users and existing instances as the new setup uses the same tools that are currently used by our customers who are deploying with the custom scripts (which use Kustomize under the hood). It is also built into kubectl, a tool that is required to perform all Kubernetes deployments.

We hope this new base cluster and deployment method will improve the user experience and provide a more industry-standard approach to deploying Sourcegraph, and more importantly, make deploying Sourcegraph easier for everyone

#### Decision flow

For users with RBAC-enabled clusters

1. Deploy all Sourcegraph services with privileged

For users without RBAC-enabled cluster

1. Deploy the new base cluster as non-root and non-privileged, which should work on any running clusters with the correct network configurations
1. If RBAC resources are allowed at a later time, they can then deploy RBAC resources for the frontend to enable service discovery
- This method also makes it easier for site admins that do not have permissions to create RBAC resources, as they can now defer to their cluster admin to deploy the services that they do not have permission to install without blocking moving to next steps with Sourcegraph

## Progress Updates / Notes

TODO:
[x] Combine current non-privilege and non-privilege-create-cluster overlays into one
[x] Replace searcher deployment and symbols deployment with StatefulSets
[x] Add `fsGroupPolicy` to replace the old `non-privileged-create-cluster` overlay
[x] Add commonly-used kustomize overlays and components for the new base cluster
[x] Add component and overlay for k3s deployment
[x] Move all monitoring-related resources into a separate directory
[x] Add component and overlay for instance size - XS
[x] Add component and overlay for instance size - S
[x] Add component and overlay for instance size - M
[x] Add component and overlay for instance size - L
[x]  Add component and overlay for instance size - XL
[x]  Test searcher and symbols as StatefulSets
[x]  Remove all rbac resources from frontend
[x] Remove rolebinding for frontend

Issues:
- [x] Empty dashboards in Grafana: update the configmap for k3s deployments fixed the issue
- [ ] Endpoints for each services must be input manually as en var to frontend: address in https://github.com/sourcegraph/sourcegraph/pull/45862

### Checklist

## How to test

Instructions to build / deploy the new cluster using the new Kustomize set ups.

1. Start docs site from this PR: https://github.com/sourcegraph/sourcegraph/pull/46227
1. Follow the instructions in the Kustomize section to build and deploy with changes created in this PR

WIP: plugin to generate endpoints at build time

```bash
# NOT READY FOR TESTING
 kustomize build --enable-alpha-plugins --enable-exec new/plugins/example -o new/generated-cluster.yaml
```


beyang:
I tried running this command (from chat):
```
kustomize build new/overlays/k3s > new/generated-cluster.yaml
```

But it appears there's no `new/overlays/k3s` directory.

abeatrix:
@beyang 

> But it appears there's no new/overlays/k3s directory.

Apologies as I've moved the k3s overlays into the `quick-start` directory. Can you please try again with the following command?

```sh
kustomize build new/overlays/quick-start/k3s/xs > new/generated-cluster.yaml
```

> Is there yet an overlay that reproduces the existing (old) cluster from the new base cluster?

I've created a new overlay called `privileged` that would reproduces the existing cluster from the new base cluster. Searcher and symbols will still be run as Statefuls though --do you think deployment should be used instead?

You can test the overlay using this command:

```sh
kustomize build new/overlays/quick-start/privileged/xs > new/generated-cluster.yaml
```

> Can you write a test plan that walks through how a reviewer should go through and validate the correctness of the changers? 

I've updated the Test plan section in this PR for generating manifests using an existing overlay as an example, please let me know if there is anything else I should add. 

> This should include running through the setup docs as a user would, including simple cases that involve only one overlay, and a more complex one that requires composing multiple overlays

Detailed instructions for using an overlay, as well as creating a complex overlay that requires composing multiple components can be found in the [README](https://sourcegraph.sourcegraph.com/-/editor?remote_url=https%3A%2F%2Fgithub.com%2Fsourcegraph%2Fdeploy-sourcegraph.git&branch=bee%2FnewBase&file=new%2FREADME.md&editor=VSCode&version=2.2.14) inside the `new` directory

beyang:
- Have you found a way to easily diff the output of the new cluster against the existing cluster? We need to have some way to check the diff between the old and new output, in order to ensure there are no gotachs in the migration process.
- We need an overlay that reproduces the old cluster with no changes, so yes, we should have an option to emit searcher and symbols as deployments, rather than statefulsets
- Docs feedback:
  - Rather than offer two ways of using the overlays without preferring one over the other, I would propose we be opinionated about which path would be best. My preference would be:
    - Have the customer fork the repository, making it clear they should only edit files in designated directories
    - Output manifest locally (`kustomize build` and then `kubectl apply -f`, not `kubectl apply -k`) and store the generated manifest in git in order to maximize auditability
    - Emit into a directory with multiple files, rather than as a single file. I'd prefer to keep prior multi-file format for greater continuity (think about minimizing the diff for folks that migrate from old to new)
    - WDYT?
- It's a bit unclear to me reading through the docs how much config the customer will be required to write.
  - If we are asking them to copy overlay files, that seems like it could be an issue down the road if we determine we need to update our overlays (e.g., to accommodate the addition of new services or other substantial changes to the base cluster).
  - I think a few example directories showing exactly what the customer needs to write would help make this more concrete for me. For example, if I wanted to combine some of the overlays (e.g., large size, set a storage class, customize replica counts, and add permissions), what does that look like from the customer's perspective?

abeatrix:
@beyang 

> Have you found a way to easily diff the output of the new cluster against the existing cluster? We need to have some way to check the diff between the old and new output, in order to ensure there are no gotachs in the migration process.

I believe we can use the kubectl built-in function to do diff the output of the new cluster against an existing cluster that is currently deployed. I will add that to the docs.

```bash
kubectl diff -f new/generated-cluster.yaml
```

> We need an overlay that reproduces the old cluster with no changes, so yes, we should have an option to emit searcher and symbols as deployments, rather than statefulsets

I've created an overlay that uses the manifests from the existing base:

```bash
kustomize build new/overlays/quick-start/old > new/generated-cluster.yaml
```

This currently does not work with other components since all the new components work with searcher and symbols as statefulsets, I can work on those next.


Will work on addressing the docs feedbacks tomorrow.

abeatrix:
Next Steps:
[ ] Update overlay that generates the existing cluster from the new cluster to use Deployments for searcher and symbols
[ ] Update all components to support both Deployments and Statefulsets for searcher and symbols
[ ] Create an overlay for migrating from existing cluster to new cluster
[ ] Docs update: migrations steps (for generating the new manifests and how to compare the new with the old)
[ ] Docs update: steps to deploy using the new kustomize overlays
[ ] Docs update: instructions to configure cluster using provided components
[ ] Docs update: how to make customized configuration
[ ] Docs update: create examples on how to create custom overlays

.buildkite/verify-yaml.sh
```diff
@@ -5,7 +5,7 @@ cd "$(dirname "${BASH_SOURCE[0]}")/.."
 
 gcloud container clusters get-credentials cloud --zone us-central1-f --project sourcegraph-dev
 
-kubectl apply --dry-run --validate --recursive -f base/ --context=gke_sourcegraph-dev_us-central1-f_cloud
+kustomize build ./base/. | kubectl apply --dry-run --validate --context=gke_sourcegraph-dev_us-central1-f_cloud -f -
```
beyang, line 8: Why the change in the CI for the base cluster? I assume this is technically equivalent to before, but I could see the motivation be that we want to test a command that is more directly what we ask users to run.

Would it also make sense to add CI commands for `new/base`?

kubectl-apply-all.sh
```diff
@@ -12,4 +12,7 @@
 #
 # Apply the base Soucegraph deployment
 # shellcheck disable=SC2068
-kubectl apply --prune -l deploy=sourcegraph -f base --recursive $@
+echo "Printing diff"
```
beyang, line 15: Do we need to change the commands for the old base cluster? These seem roughly equivalent, but if that's the case, I would prefer not changing them in case there are subtle differences or if changing this file causes a merge conflict for anyone who has forked this repository.

.buildkite/verify-yaml.sh
abeatrix: > Why the change in the CI for the base cluster? 

this is due to the[ kustomization files added to the original base](https://github.com/sourcegraph/deploy-sourcegraph/tree/bee/newBase/base) that requires kustomize to run, but since we will be setting up a separate repo for the new base, this change will not be necessary.

> Would it also make sense to add CI commands for new/base?

It makes sense to me! I've created a new issue for creating tests for the new base https://github.com/orgs/sourcegraph/projects/290?pane=issue&itemId=18056795

abeatrix:
Moved to https://github.com/sourcegraph/deploy-sourcegraph-k8s

PR #4224, opened 2023-01-24
executor: deploy on k8s

davejrt:
This is my interpretation of https://github.com/sourcegraph/sourcegraph/issues/42469

It's a WIP as there is no documentation but we need an artefact for folks to deploy as well, and before I can document the process, we need to agree on what that is. 

Right now this creates a single replica, with yet to be determined resource requirements. It deploys a `docker in docker` sidecar to avoid exposing the underlying container runtime on the k8s host. It also deploys a pull through cache to avoid docker rate limiting. 

Assuming this is good, further steps are to ensure this works with kustomize/helm and is documented with a readme and exisiting executor docs are amended. 

### Checklist

### Test plan

Tested running a SSBC on S2

image

### DIFF
--- a/configure/executor/docker-daemon.ConfigMap.yaml
+++ b/configure/executor/docker-daemon.ConfigMap.yaml
@@ -1,0 +1,11 @@
+apiVersion: v1
+data:
+  daemon.json: |
+    { "insecure-registries":["private-docker-registry:5000"] }
+
+kind: ConfigMap
+metadata:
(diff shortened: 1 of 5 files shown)


eseliger (commented):
this looks pretty solid to me :)

eseliger:
using a separate docker daemon actually seems really smart to work around the issues, great idea!

PR #4231, opened 2023-03-14
executors: fix bugs in manifests

davejrt:
This PR fixes an issue with the metadata on the private docker registry. It also adds in an updated image with NO cve's and limits the scratch storage to avoid the pod being evicted for causing disk pressure based on the docs. 

It also fixes the liveness/readiness probes for the executor. 

### Checklist

### Test plan

Tested on dogfood.

### DIFF
--- a/configure/executors/executor/executor.Deployment.yaml
+++ b/configure/executors/executor/executor.Deployment.yaml
@@ -31,15 +31,18 @@ spec:
           image: index.docker.io/sourcegraph/executor:insiders@sha256:deda81624796df31b56bdaa6820b3332eaa2cbe5e1729886d15ce3c6f57b69b5
           imagePullPolicy: Always
           livenessProbe:
-            command:
-              - /usr/local/bin/executor
-              - validate
+            exec:
+              command:
+                - /usr/local/bin/executor
+                - validate
             initialDelaySeconds: 60
-            timeoutSeconds: 5
+            timeoutSeconds: 30
           readinessProbe:
-            command:
-              - /usr/local/bin/executor
-              - validate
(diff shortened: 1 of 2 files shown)


configure/executors/executor/executor.Deployment.yaml
```diff
@@ -28,25 +28,21 @@ spec:
...
-              scheme: HTTP
-            periodSeconds: 5
-            timeoutSeconds: 5
-          ports:
-            - containerPort: 6060
-              name: debug
+            exec:
+              command:
+                - /usr/local/bin/executor
+                - validate
```
eseliger, line 44: if sourcegraph is down, this would also kill executors _I think_, would that be ok?

eseliger (approved):

configure/executors/executor/executor.Deployment.yaml
davejrt: Bit of a chicken and egg scenario, if SG is down presumably this isn't working anyway. And otherwise there isn't really a nice way to see if this is ready for work. `pgrep <large tini command>` is ugly and we don't open a port. The fact that if the frontend for example is down...we also lose gitserver,repo-updater and a host of other services via readiness checks was my reasoning here. 

Thoughts?

eseliger: Yeah, no strong feelings on this, I don't _love_ that a frontend outage would mean other things cascadingly also go down fully, but nbd either.

davejrt: news flash, found a better pgrep command

PR #4272, opened 2023-10-03
all: refresh image tags

unknwon:
### Description

1. Skip broken test suite
2. Refresh all image tags, and some are also using non-`insiders` that somehow breaks the update image tag script, resulting us releasing very old image versions
3. Fix an error check bug

### Test plan

CI


unknwon:
Merging as-is to unblock the release. Feel free to do post-merge reviews!

PR #4288, opened 2024-01-04
update tools go.mod and fix package import

burmudar:
### Fixes
* ci `check-image-names.sh` was failing due to a package import change. In the process of updating the import I also updated the go.mod to be 1.21
* fix `.buildkite/prettier-check.sh` by moving package.json to the root of the repo. With it being in root of the repo it won't install the integration test dependencies to run prettier. 

### Test plan
Executed locally and CI


go.mod
```diff
@@ -1,20 +1,17 @@
-module github.com/sourcegraph/deploy-sourcegraph
+module sourcegraph/deploy-sourcegraph
 
-go 1.16
+go 1.21.5
```
Strum355, line 3: Does this need bumping too? https://github.com/sourcegraph/deploy-sourcegraph/blob/master/.github/workflows/update-tags.yml#L17

burmudar: Ah good catch. Might as well update it!

Strum355 (approved):

jhchabran (approved):

